PR #31, opened 2013-04-28
Add #go method for chainable navigation

mattvanhorn:
The new #go method allows a user to load the page
and interact with it in a single line step body.

Instead of:
my_page.load
my_page.do_something

You can do:
my_page.go.do_something

### DIFF
--- a/features/messages.feature
+++ b/features/messages.feature
@@ -1,0 +1,10 @@
+Feature: Page navigation
+  As a tester
+  I want to be able to use abstractions for messages on the page
+  In order to verify the page
+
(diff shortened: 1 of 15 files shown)


vmpj:
I like this feature, but maybe accepting a block would be more useful. E.g.

    Given /^I am logged into Commons$/ do
      visit Commons::LoginPage do |page|
        page.login_with Commons::USERNAME, Commons::PASSWORD
        page.should be_logged_in
      end
    end

See #100

natritmeyer:
Tidying up ancient pull request

PR #67, opened 2014-03-25
allow block syntax on sections

tgaff:
This makes for shorter test code when working within nested sections.
 - similar to Capybara's within

Compare:
```ruby
@page.top_section.child_section.should have_element1
@page.top_section.child_section.should have_element2
@page.top_section.child_section.should have_element3
````

```ruby
@page.top_section.child_section.within do |child|
  child.should have_element1
  child.should have_element2
  child.should have_element3
end
````

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,9 +7,6 @@ Documentation:
 AccessorMethodName:
   Enabled: false
 
-Style/SymbolProc:
-  Enabled: false
-
 TrivialAccessors:
   Enabled: false
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,6 +7,9 @@ Documentation:
 AccessorMethodName:
   Enabled: false
 
+Style/SymbolProc:
+  Enabled: false
+
 TrivialAccessors:
   Enabled: false
 
--- a/README.md
+++ b/README.md
@@ -868,10 +868,10 @@ Some of this test code can be made a little prettier by simply passing a block i
 ```ruby
 Then /^the home page menu contains a link to the various search functions$/ do
   @home.menu do |menu|
-    menu.should have_search
-    menu.search['href'].should include "google.com"
-    menu.should have_images
-    menu.should have_maps
+    expect(menu).to have_search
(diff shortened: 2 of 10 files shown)


tgaff:
Hi, @natritmeyer  What are your thoughts on this?  Is there interest in getting it in?  Should I kill it?

natritmeyer:
I've been flat out on other projects recently, but I'm free to work on site_prism for a while :) I'll take a look over the coming days.

tgaff:
re-re-rebased. 

Note that I still expect this to fail on rubocop until #121 is merged and this is rebased once more.

jnyman:
@tgaff Just checking something: it's not clear to me from the comments -- is the final code going to work with iframes as well?

I can say that the bit of code shown here:

    @page.within(:menu) do |menu|
        expect(menu).to have_search
    end

Makes as lot of sense to me, particularly if that :menu was an iframe. When discussing or testing anything with iframes, people do tend to discuss elements as being "within" the iframe. This is probably why Capybara uses this type of wording.

andrewhao:
I'd love to help verify and bring this PR into reality. What can I do to help?

tgaff:
@jnyman sorry for the late reply.  This change doesn't explicitly deal with iframes because site-prism already has support for that in a very similar way to how you show it.   

Instead this works with SitePrism Sections which are a little like partials for your page-objects.  You'll be able to work with sections the same way you can with iframes already.

tgaff:
@andrewhao Hi!  If you happen to be using this PR in your code already that would be great to know!  If you're not, it'd be great if you gave it a try and let us know if you run into any issues.  If you want take a look at the code/tests too.

At some point I believe Nat had indicated he was considering merging this.  That was quite a long time ago though and I know he's had good reason to be busy of late.  I should reach out to him about it though.

natritmeyer:
Well, that took a while :/ Apologies!

natritmeyer:
It's probably going to fail in travis due to this: https://github.com/rnelson0/puppet-local_user/pull/44 But that's just rubocop. The PR works fine :)

PR #86, opened 2014-07-14
Add ability to attach sections without explicitely set section's ClassName

airatshigapov:
Current realisation doesn't allow sections to be attached without explicit class name set, for example we are forced to use this:

```ruby
class LoginPage < SitePrism::Page
  section :top_bar, TopBar, '.top-bar'
end
```

when we can just use this:

```ruby
class LoginPage < SitePrism::Page
  section :top_bar, '.top-bar'
end
```

`TopBar` class name generates automatically with `camelize` method. This also doesn't break existing behaviour and generates class name from section name only when no explicit class name provided.


airatshigapov:
Almost 2 years have passed and I've lost original fork and branch which were used for this PR, so I'm closing this one :)

Renovated PR is here https://github.com/natritmeyer/site_prism/pull/161

PR #93, opened 2014-11-04
Add examples using label text to fined elements.

MarkBorcherding:
We found it really useful to use form labels as the selector for finding inputs. There were no examples of this in the Readme, but it works since the parameters are delegated down to Capybara. 

I would consider submitting a PR to add those helpers to `SitePrism::Page` if there is a chance it would be accepted. I'm not sure if that additional API is desired.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,6 +362,30 @@ Here we're adding a search field to the Home page. The `element` method
 takes 2 arguments: the name of the element as a symbol, and a css selector
 as a string.
 
+#### Finding elements via text
+
+An element can be found via its name, id, text, label text, or value, depending
+on the type of element.
+
+```ruby
+class Login < SitePrism::Page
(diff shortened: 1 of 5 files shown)


natritmeyer:
Sorry I've taken a while to get to this...

Isn't this just normal capybara functionality? SitePrism will just pass any parameters down to capybara, so not sure what we're getting here...

Would an addition to the readme be enough? You're right that some people would use these features if they knew they existed :)

MarkBorcherding:
You're right that the readme would technically be enough. When I didn't see it mentioned I started looking in the specs as a second form of documentation. I didn't see it mentioned there either so I added it to make sure it would work. 

I might suggest if it is going to be mentioned in the readme as a valid use that there should be a test supporting that claim.

PR #98, opened 2014-12-19
Allow for excluded elements in a page's all_there? check

TheMetalCode:
Let's say the page I'm modeling with a SitePrism::Page class has at least one element that isn't always guaranteed to be present.

``` ruby
  element :name_field, '#name'
  element :err_msg, '.error'
end
```

If I ever wished to leverage that page's all_there? check, it would fail because the err_msg would not be present on initial load (hopefully). And yet, I would very much like to use the page's all_there? check.

This pull request adds the ability for any page class to define an excluded_elements method, which would return an array of elements to exclude from the all_there? check.


lib/site_prism/element_checker.rb
```diff
@@ -1,9 +1,35 @@
...
   module ElementChecker
+
     def all_there?
       Capybara.using_wait_time(0) do
-        self.class.mapped_items.all? { |element| send "has_#{element}?" }
+        expected_elements.all? { |element| send "has_#{element}?" }
       end
     end
+
+    def excluded_all_absent?
```
TheMetalCode: I did not end up using this method a whole lot, but I thought there might be at least some cases where one would want to test whether or not the elements in excluded_elements are actually absent.

luke-hill:
This is one of a few PR's raised which I'm looking into, and just generally giving a "bump".

Could you provide an update if at all possible into what you want to do with this PR (I'm aware it's been left a while)

Thanks for your contribution.

Note: I noticed you mentioned you think this is duplicate code. If so lets just close this PR

TheMetalCode:
@luke-hill Thanks for reaching out! Since submitting this PR I moved away from using SitePrism, and am now in another job where we do use SitePrism again, so it's actually a good time for me to dig back in and see if I still think this code is useful. I'll update this weekend for sure.

Also, my fault for not clarifying: the duplication comment was in reference to solving rubocop errors that were on my branch that were being addressed in another PR, and not the code to allow for excluded elements in an `all_there?` check.

TheMetalCode:
https://travis-ci.org/natritmeyer/site_prism/builds/342765313?utm_source=github_status&utm_medium=notification

^^ Demonstrates that when Rubocop tech debt is excluded and geckodriver is added to the travis config (see https://github.com/natritmeyer/site_prism/pull/186/files), this build would likely pass.

TheMetalCode:
@luke-hill OK, given a merge-in of latest master and a quick brush-up, I'd still like for this PR to be considered. The `all_there?` check is something I tend to want to use on initial page load to avoid having to use a lot of specific `wait_until_some_element_visible` statements later on, and the ability to exclude elements not always guaranteed to be present on a given page seems useful to me. 

I did rewrite the PR description a bit, because I think the modal example wasn't really the best illustration - I'd probably use distinct page object classes and not sections for those anyways.

luke-hill:
I'm happy for this code as the code suggests. However (Not yet), at some point next week I'll probably review with some stylistic stuff. However the code as is does enough to warrant a merge without massive refactors.

I'll get round to this at some point in a week or so, this code is unlikely to hit 2.9.x or a 2.10 candidate. But it will definitely hit a release soon after. 

Thanks for your support.

Luke

README.md
```diff
@@ -706,6 +706,19 @@ end
 
 ```
 
+You may wish to have elements declared in a page object class that are not always guaranteed to be present (success or error messages, etc.). If you'd still like to test such a page with `all_there?` you can define an `excluded_elements` method on your page object class that excludes the specified elements from the `all_there?` check.
+
+```ruby
+class TestPage < SitePrism::Page
+  element :name_field, '#name'
```
luke-hill: Optional - I'm going to change the docs at some point to use `()` instead of spaces as it's more declarative and avoids any issues. This is going to be a suite wide linting change. If you want to tweak yours here that would be nice.

features/page_element_interaction.feature
```diff
@@ -44,6 +44,10 @@ Feature: Page element interaction
     When I navigate to the home page
     Then all expected elements are present
 
+  Scenario: Page with excluded_elements
```
luke-hill: Optional Linting
As you're directly referencing a method inside a Scenario wrap it in backticks (Key left of 1), to show this

features/page_element_interaction.feature
```diff
@@ -44,6 +44,10 @@ Feature: Page element interaction
...
+    When I navigate to the home page with excluded elements
```
luke-hill: `home page that contains excluded elements`

lib/site_prism/element_checker.rb
```diff
@@ -2,8 +2,18 @@ module SitePrism
...
-        self.class.mapped_items.all? { |element| send "has_#{element}?" }
+        elements_to_check.all? { |element| send "has_#{element}?" }
       end
     end
+
+    private
+
+    def elements_to_check
+      elements = self.class.mapped_items
+      if self.respond_to? :excluded_elements
```
luke-hill: wrap args in `()` and el after include on line below

spec/element_spec.rb
```diff
@@ -45,6 +45,23 @@ class PageWithAFewElements < SitePrism::Page
     expect(page).to respond_to :all_there?
   end
 
+  it 'checks only expected_elements if excluded_elements specified' do
+    class PageWithAFewElements < SitePrism::Page
+      element :bob, 'a.b c.d'
```
luke-hill: Same as description on README, up to you though as I know it's in keeping style.

spec/element_spec.rb
```diff
@@ -45,6 +45,23 @@ class PageWithAFewElements < SitePrism::Page
...
+      element :success_msg, 'span.alert-success'
+
+      def excluded_elements
+        %w[success_msg]
+      end
+    end
+    page = PageWithAFewElements.new
+    allow(page).to receive(:has_bob?).and_return(true)
+    allow(page).to receive(:has_success_msg?).and_return(false)
+    expect(page.all_there?).to be_truthy
```
luke-hill: Originally wrote this was redundant, but you're making the call here.

Perhaps instead just verbosely call `page.all_there?` I know the whole stylistic area of our Specs needs a lot of work (RE Let blocks and formatting)

test_site/pages/home_with_excluded_elements.rb
```diff
@@ -0,0 +1,42 @@
...
+  element :go_button, :xpath, '//input'
+  element :link_to_search_page, :xpath, '//a'
+  element :some_slow_element, :xpath, '//a[@class="slow"]'
+  element :invisible_element, 'input.invisible'
+  element :shy_element, 'input#will_become_visible'
+  element :retiring_element, 'input#will_become_invisible'
+  element :remove_container_with_element_btn, 'input#remove_container_with_element'
+
+  # elements groups
+  elements :lots_of_links, :xpath, '//td//a'
```
luke-hill: I don't generically have an issue with what you've done here to obtain Test Isolation. However I do believe that this class could be 25-50% lighter. 

To start with lets remove all plural items (`sections` and `elements`), and perhaps just have a couple of `element` and `section` objects. With of course the element objects that don't exist.

2.10 / 2.11 are going to be massively trimmed down dieted versions ideally of SitePrism so it's easier to manage going forwards.

spec/element_spec.rb
TheMetalCode: In my mind, `expect(page.all_there?).to be_truthy` is a fairly key assertion, perhaps even more important than the two below it, because the idea behind `excluded_elements` is to allow for `all_there?` to return true even when elements specified ahead of time are not present on the page. It just happens to also function as a call to `all_there?` that enables the below two assertions about the stubbed methods.

Arguably this could be two specs, under a `describe` block. That's a good area of improvement for the entire rspec suite, along with wider use of `let` blocks and other matters.

test_site/pages/home_with_excluded_elements.rb
TheMetalCode: The more I think about it, the more I'd like to actually have a pared down htm file to match a pared-down page object class to test the `excluded_elements` case. Will add that as well.

README.md
```diff
@@ -711,6 +711,19 @@ end
 
 ```
 
+You may wish to have elements declared in a page object class that are not always guaranteed to be present (success or error messages, etc.). If you'd still like to test such a page with `all_there?` you can define an `excluded_elements` method on your page object class that excludes the specified elements from the `all_there?` check.
+
+```ruby
+class TestPage < SitePrism::Page
+  element(:name_field, '#name')
```
tgaff: The style in (most of?) the rest of the documentation does not use parentheses around these arguments.  If you have time maybe change it to be consistent.

e.g. `  element :search_field, "input[name='q']"`

README.md
```diff
@@ -711,6 +711,19 @@ end
...
+  element(:success_msg, 'span.alert-success')
+
+  def excluded_elements
```
tgaff: Why not just have this as an array?  Why a method?

test_site/html/home_with_excluded_elements.htm
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Home Page With Excluded Elements</title>
+    <script type="text/javascript">
+    function removeElement(node) {
```
tgaff: I think this script and some of the html content below is unnecessary.  The next developer in here may not be aware of what is and isn't actually used, so it's best to keep it minimal.

test_site/pages/home_with_excluded_elements.rb
```diff
@@ -0,0 +1,16 @@
+class TestHomePageWithExcludedElements < SitePrism::Page
+  set_url('/home_with_excluded_elements.htm')
+  set_url_matcher(/home_with_excluded_elements\.htm$/)
+
+  # individual elements
+  element(:welcome_header, :xpath, '//h1')
+  element(:welcome_message, :xpath, '//span')
+  element(:go_button, :xpath, '//input')
```
tgaff: Similar to above, I don't think most of this is used, right?  
2 or 3 elements is probably sufficient.

README.md
TheMetalCode: That conflicts with prior review input from @luke-hill that explicitly asks for the parentheses to be used.

README.md
tgaff: Oh... now I see this.  
I'm afraid that I just suggested the opposite.  😓 Sorry for the confusion @TheMetalCode 

@luke-hill I'm not sure why you want to add the parentheses, but would it make more sense to just do them all at once rather than piece-meal?
The ruby style guide suggests that we should omit parentheses for "Methods that are part of an internal DSL (e.g., Rake, Rails, RSpec)".  I think this applies.
https://github.com/bbatsov/ruby-style-guide#syntax

README.md
TheMetalCode: I'm not sure I understand the question - do you mean why a method instead of a class constant or an instance variable? Given that the logic I added in lib/site_prism/element_checker.rb is meant to operate on `excluded_elements` as some kind of member of a page object class, I'm not sure what having this be a simple array would mean. It has to be either some kind of member variable that holds an array, or what I have here, a method that returns an array.

README.md
tgaff: Right, my bad.  I didn't know he was trying to change that.  I commented his comment.

README.md
tgaff: What I'm trying to say is that this is the only point in an otherwise declarative DSL where the user is instructed that they need to define a method, so it maybe feels a little out of place.


```ruby
  element(:name_field, '#name')
  element(:success_msg, 'span.alert-success')

  def excluded_elements
    %w[success_msg]
  end
end
```

I'm wondering if you considered something like the below implementation:

```ruby
  element(:name_field, '#name')
  element(:success_msg, 'span.alert-success')

  excluded_elements ['success_msg', 'alert']
 
end
```

tgaff:
You could even take this a step further:

```ruby
  element(:name_field, '#name')
  element(:success_msg, 'span.alert-success')
  element(:login_button, '#login')
  excluded_elements %w[success_msg alert]

  def logged_in_exclusions
    ['login_button']
  end    

end

@page.all_there? => true
@page.all_there?(except: :name_field) => true
@page.all_there?(except: @page.logged_in_exclusions)  # ok, maybe not the prettiest implementation
```

README.md
TheMetalCode: Gotcha. Yeah, that is a fair point, keeping up with the declarative DSL probably is going to flow a bit better. I'll give that a spin. Thanks!

test_site/html/home_with_excluded_elements.htm
TheMetalCode: Yeah, this is kept from the `home.html` as something to enable an element that isn't immediately present but becomes present later. Which...isn't _strictly_ necessary for the purpose of this test so I'll pare down for sure.

TheMetalCode:
@tgaff Food for thought - thanks! I think I like the idea of `except` as an optional arg to `all_there`.

README.md
TheMetalCode: Hmm...I guess I'd say if Rubocop isn't going to enforce parentheses for things like DSL declarations then I'm having a hard time coming up with a reason why we should. My two cents.

luke-hill:
Given I had already signed this off (And at the time wasn't getting reviews from others). What do we want to do.

I think as is it is perfectly fine to hit a `2.11` candidate (Tagged). And can always be iterated on.

I'd prefer things such as test cleanup e.t.c. if done separately to look at all situations. And I will be looking into bits and pieces of this going forward on the `TODO.md` doc.

@TheMetalCode Do you want to make changes in this PR. Or raise a new PR in 2.11 once this is merged in (Either option is fine by me). @tgaff Are you happy with this also?

README.md
luke-hill: I've had a read of that article. I was aware of it being the case for things like Rails e.t.c. I guess it's the case that do we declare this is our internal DSL. In which case we probably would. SO happy to be proven wrong here.

There are plenty of instances elsewhere though where we don't use `()` and I will be switching those over. But happy to admit I'm wrong here. 

For reference, at my company we use `()` btu I'm happy for the declared style to omit them. If you both feel as though that is best.

TheMetalCode:
@luke-hill I think I'd like to shoot for one change in this PR, addressing @tgaff's suggestion to make `excluded_elements` part of the DSL (like `element`) and not just a method, as I think that would be a significant improvement that I'd like to get right the first time. I think everything else can be iterated on later.

README.md
TheMetalCode: I'd probably prefer to leave out the parentheses for the DSL declarations but enforce them for plain-old ruby method arguments and other places where we'd expect them to be present.

tgaff:
Sounds good to me!  Let me know if I can help (although I'm not free again until Wed)!

TheMetalCode:
@luke-hill @tgaff I gave this some further thought and would like to propose https://github.com/natritmeyer/site_prism/pull/208 as an alternative to this PR. I think `expected_elements` might be a better approach than `excluded_elements`.

luke-hill:
Oki. Lets close this PR then and mark that as ready for review.

PR #108, opened 2015-03-30
Update the system to more carefully combine arguments.

RealNobody:
I like the options to pass parameters into the finders.

This allows me to create a definition like this:

```
class MyPage < SitePrism::Page
  elements :my_elements, ".my-class"
end
```

Then, if I have a `MyPage` object I can do something like:

```
my_page.my_elements(minimum: 12) # only return success if there are at least 12 items.
```

The problem I ran into is if I had the following definitions:

```
class MyPage < SitePrism::Page
  elements :my_visible_elements, ".my-class"
  elements :my_possible_elements, ".my-class", visible: false
end
```

Now, if I make this call, I don't get my expected results:

```
my_page.my_possible_elements(minimum: 12) # not quite right - will be the same as my_page.my_visible_elements(minimum: 12)
```

Instead, I would have to do:

```
my_page.my_possible_elements(minimum: 12, visible: false)
```

But this means that I have to know what the options were that were a part of the element definition.

This change merges the sets of values in the element definition with the values passed in to the defined object such that there is only one options hash that is a merge of the two.

This would now allow:

```
my_page.my_possible_elements(minimum: 12) # now, the options is { visible: false, minimum: 12 }
```

### DIFF
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -6,7 +6,7 @@ module SitePrism
       build element_name, *find_args do
         define_method element_name.to_s do |*runtime_args, &element_block|
           self.class.raise_if_block(self, element_name.to_s, !element_block.nil?)
-          find_first(*combine_args(find_args, runtime_args))
+          find_first(*self.class.send(:combine_args, find_args, runtime_args))
         end
       end
     end
@@ -15,7 +15,7 @@ module SitePrism
       build collection_name, *find_args do
         define_method collection_name.to_s do |*runtime_args, &element_block|
           self.class.raise_if_block(self, collection_name.to_s, !element_block.nil?)
-          find_all(*combine_args(find_args, runtime_args))
+          find_all(*self.class.send(:combine_args, find_args, runtime_args))
         end
       end
     end
@@ -26,7 +26,7 @@ module SitePrism
       build section_name, *find_args do
         define_method section_name do |*runtime_args, &element_block|
           self.class.raise_if_block(self, section_name.to_s, !element_block.nil?)
-          section_class.new self, find_first(*combine_args(find_args, runtime_args))
+          section_class.new self, find_first(*self.class.send(:combine_args, find_args, runtime_args))
         end
       end
     end
@@ -36,7 +36,7 @@ module SitePrism
       build section_collection_name, *find_args do
         define_method section_collection_name do |*runtime_args, &element_block|
           self.class.raise_if_block(self, section_collection_name.to_s, !element_block.nil?)
-          find_all(*combine_args(find_args, runtime_args)).map do |element|
+          find_all(*self.class.send(:combine_args, find_args, runtime_args)).map do |element|
             section_class.new self, element
           end
         end
@@ -112,7 +112,7 @@ module SitePrism
         define_method method_name do |*runtime_args|
           wait_time = SitePrism.use_implicit_waits ? Capybara.default_wait_time : 0
           Capybara.using_wait_time wait_time do
-            element_exists?(*combine_args(find_args, runtime_args))
+            element_exists?(*self.class.send(:combine_args, find_args, runtime_args))
           end
         end
       end
@@ -124,7 +124,7 @@ module SitePrism
         define_method method_name do |*runtime_args|
           wait_time = SitePrism.use_implicit_waits ? Capybara.default_wait_time : 0
           Capybara.using_wait_time wait_time do
-            element_does_not_exist?(*combine_args(find_args, runtime_args))
+            element_does_not_exist?(*self.class.send(:combine_args, find_args, runtime_args))
           end
         end
       end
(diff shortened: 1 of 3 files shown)


spec/combine_args_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let(:array_options_args1) do
+    array = rand(1..5).times.map { rand(50..200).times.map { ('a'..'z').to_a.sample }.join("") }
+    array << rand(1..5).times.reduce({}) { |hash, _| hash[rand(5..20).times.map { ('a'..'z').to_a.sample }.join("")] = rand(50..200).times.map { ('a'..'z').to_a.sample }.join(""); hash }
+    array
+  end
+  let(:array_options_args2) do
+    array = rand(1..5).times.map { rand(50..200).times.map { ('a'..'z').to_a.sample }.join("") }
+    array << rand(1..5).times.reduce({}) { |hash, _| hash[rand(5..20).times.map { ('a'..'z').to_a.sample }.join("")] = rand(50..200).times.map { ('a'..'z').to_a.sample }.join(""); hash }
+    array
+  end
```
tmertens: Use static array options in these let blocks, and hardcoded expected values in the tests (if the same value is used across multiple tests, it is OK to memoize it in a let() block but do not compute the values if at all possible).

These let blocks overcomplicate the tests and also force you to duplicate the production logic in the tests themselves to create the expected result arrays.  Computed expectation values are also not great from a documentation standpoint because you have to run the test and output the resulting expectation to see what it is supposed to be, or read through all of the test/setup code to figure out what the result should be, rather than simply looking at the test and seeing a hardcoded result value.  Computed expectations also make the tests more brittle, difficult to maintain, and more likely that something may break and go undetected.

lib/site_prism/element_container.rb
```diff
@@ -69,6 +69,18 @@ def raise_if_block(obj, name, has_block)
 
     private
 
+    def combine_args(arg_array1, arg_array2)
+      args = []
+      args += arg_array1
```
tmertens: ```args = arg_array1.dup```

lib/site_prism/element_container.rb
```diff
@@ -69,6 +69,18 @@ def raise_if_block(obj, name, has_block)
...
+      options1 = (args.pop if args[-1].is_a?(Hash)) || {}
```
tmertens: maybe we should refactor the arguments to their own class?  I wouldn't say it's necessary for this PR, that's up to @natritmeyer but it seems like this method and the logic within is not the responsibility of an ElementContainer.  It is probably worth a refactor at some point.

tmertens:
This is great, @RealNobody!

I've been meaning to do this for some time, it was a known/accepted flaw of the original implementation of runtime arguments so I'm glad to see it finally being resolved.  I would simplify the tests (see my comment) but otherwise this looks great.

spec/combine_args_spec.rb
RealNobody: OK

lib/site_prism/element_container.rb
RealNobody: > It is probably worth a refactor at some point.

I wasn't sure where to put this function really.  This seemed semi-reasonable at the time.

RealNobody:
> it was a known/accepted flaw

This implementation also has a minor "flaw" in that it always passes in options to the find functions.  That is if I had the definition `element :my_element, "#my_element_definition_param"`, instead of calling:  `find("#my_element_definition_param")`, it will call `find("#my_element_definition_param", {})`  (it always adds a blank `{}` options hash param).

Since we're "only" calling the Capybara finder functions, I know that this isn't a problem.  If I were building this more generically, I would normally put in some extra code to ensure that we don't pass a hash unless one of the two original arrays had a hash.

For this use, it didn't seem to really be necessary.

RealNobody:
> I would simplify the tests

I tend to over-complicate tests.  One of my many flaws.

tmertens:
@RealNobody can you resolve the merge conflicts here?

FLemon:
@RealNobody @tmertens any update on this PR? am waiting to have this fix merge in, thank you

RealNobody:
> any update on this PR? am waiting to have this fix merge in, thank you

Sorry, I lost track of this.  I'll see if I can resolve now.

RealNobody:
@tmertens, @FLemon  I rebased (finally) and resolved conflicts, but the continuous integration failed, and I don't know why since I don't think I touched anything in that area...

luke-hill:
Thankyou for your contribution.

In this instance I think the easiest thing for you to do is Close this PR, Re-open a new PR based off the latest master and maybe cherry-pick one or two commits. At least then we'll have a cleaner implementation.

Don't worry about rubocop issues too much just yet, there will be a separate set of tasks to look at those.

If you're ok with doing that then thankyou for your help.

RealNobody:
Closing per-recommendation.

Hope to re-open soonish.

Thanks.

PR #117, opened 2015-07-08
Simplifies check for URI scheme

benlovell:


lib/site_prism/page.rb
```diff
@@ -73,7 +73,7 @@ def url_matcher
     end
 
     def secure?
-      !current_url.match(/^https/).nil?
+      current_url =~ /^https/
     end
```
DouglasMeyer: @benlovell Definitely an improvement as far as readability. I may be getting picky that it wouldn't return a boolean anymore. What do you think of `current_url.start_with? 'https'`?

benlovell:
@DouglasMeyer Good call. Done.

natritmeyer:
Thanks :)

Issue #120, opened 2015-08-07
rubocop is run twice (per ruby) on travis-ci


tgaff:
```
0.01s$ sh -e /etc/init.d/xvfb start
Starting virtual X frame buffer: Xvfb.
4.97s$ bundle exec rubocop
Inspecting 42 files
..........................................
42 files inspected, no offenses detected
The command "bundle exec rubocop" exited with 0.
46.01s$ bundle exec rake
Running RuboCop...
Inspecting 42 files
..........................................
42 files inspected, no offenses detected
```

https://travis-ci.org/natritmeyer/site_prism/jobs/70135720

luke-hill:
Closing as under review

Issue #119, opened 2015-08-07
latest rubocop reports 6 failures on master 


tgaff:
Master at 609e981d5feb6255dbd97ef560c18f1f9df9dcb7 is reporting 6 rubocop offenses:

```
Inspecting 42 files
.C...........C....C.......................

Offenses:

Rakefile:28:12: C: Unnecessary spacing detected.
    t.files   = ['lib/**/*.rb', '-', 'README.md']
           ^^
lib/site_prism/exceptions.rb:9:42: C: Unnecessary spacing detected.
  class UnsupportedBlock < StandardError;  end
                                         ^
lib/site_prism.rb:5:30: C: Unnecessary spacing detected.
  autoload :ElementContainer,  'site_prism/element_container'
                             ^
lib/site_prism.rb:6:28: C: Unnecessary spacing detected.
  autoload :ElementChecker,  'site_prism/element_checker'
                           ^
lib/site_prism.rb:7:18: C: Unnecessary spacing detected.
  autoload :Page,  'site_prism/page'
                 ^
lib/site_prism.rb:8:21: C: Unnecessary spacing detected.
  autoload :Section,  'site_prism/section'
                    ^

42 files inspected, 6 offenses detected
```

travis.ci reports similar failures on multiple ruby versions

Darhazer:
With 0.46 on master (currently at 307d19e70742da540c0aa337669fa8bbadf9fc56)

```
Offenses:

Gemfile:11:3: C: Gem redcarpet should appear before yard in their gem group.
  gem 'redcarpet'
  ^^^^^^^^^^^^^^^
Gemfile:13:3: C: Gem rubocop should appear before simplecov in their gem group.
  gem 'rubocop'
  ^^^^^^^^^^^^^
spec/loadable_spec.rb:19:9: C: Use normalcase for variable numbers.
        validation_1 = -> { true }
        ^^^^^^^^^^^^
spec/loadable_spec.rb:20:9: C: Use normalcase for variable numbers.
        validation_2 = -> { true }
        ^^^^^^^^^^^^
spec/loadable_spec.rb:21:9: C: Use normalcase for variable numbers.
        validation_3 = -> { true }
        ^^^^^^^^^^^^
spec/loadable_spec.rb:22:9: C: Use normalcase for variable numbers.
        validation_4 = -> { true }
        ^^^^^^^^^^^^
spec/loadable_spec.rb:78:7: C: Use normalcase for variable numbers.
[...]

PR #121, opened 2015-08-07
Fixing two seperate rubocop issues

tgaff:
Latest rubocop is unhappy with double spaces. [Completes #119]
Rubocop is run twice on travis-ci because it's called by the rake task [Completes #120]


natritmeyer:
Thanks :)

PR #125, opened 2015-09-07
Change Capybara.default_wait_time to Capybara.default_max_wait_time to fix deprecation warnings

tpbowden:
The warning appears whilst running Cucumber with recent versions of Capybara. Also added some tidying up via Rubocop


natritmeyer:
Thanks :)

PR #127, opened 2015-09-11
updated to support Capybara 2.5.0

mnohai-mdsol:
- Gets rid of deprecation warning when using implicit waits
- Supports Capybara 2.5.0


natritmeyer:
Thanks :)

Issue #50, opened 2013-10-04
all_there? does not use optional implicit waits


tmertens:
Presently, all_there? is not implemented to use the new capybara implicit waits.

luke-hill:
Closing in favour of PR/s

Issue #49, opened 2013-10-04
all_there? should be recursive


tmertens:
Presently all_there? only checks if the direct child elements of a page or section are present on the page, but does not ensure that all required descendant elements of those child elements are also present.

luke-hill:
Hi @tmertens You'll notice I've been bumping a few tickets recently and getting peoples input.

Can this and #48 be looked at and closed. I believe there is a PR tidying up `all_there?` slightly which will be merged soon.

luke-hill:
Will see if the demand for this is high enough. Have refactored this area recently though

Issue #48, opened 2013-10-04
all_there? assumes all subelements are always present


tmertens:
The all_there? method expects all elements in a container to always be present.  However, there are often page elements and entire sections which are displayed on the page depending on some condition or action of the user.

Examples:
- Error Message(s)
- Page sections which appear/disappear based on user interaction.
- Modal (Lightbox) Windows
- All descendant sections/elements of aforementioned 'sections'.

It is necessary to have a way of flagging certain elements as optional so that they and any of their descendants do not cause the "all_there?" check against an ancestor element to fail.

Given the page:

``` ruby
class LoginPage < SitePrism::Page
  class LoginForm < SitePrism::Section
    element :password_field, '#password'
    element :username_field, '#username'
    element :submit_button, 'button[type="submit"]'
  end

  class NoticeWindow < SitePrism::Section
    element :notice, ".notice"
    element :alert, ".alert"
  end

  set_url "/login"
  set_url_matcher /login/

  # Page content
  section :login_form, LoginForm, "#loginform"
  element :logo, 'img.logo'
  section :notice_window, NoticeWindow, ".notice-window"
end
```

It is obvious that login_form and logo should be present on page load, and that password_field, username_field, and submit_button should also be present; however, notice_window and/or its subelements should only be present/displayed when a warning or notice must be displayed on the page such as after an invalid login.

So, presuming the required fields described above are present on page load, the following should succeed, but does not:

``` ruby
page = LoginPage.new
page.load
page.should be_all_there #=> all_there? returns false and fails validation
```

Do you have any thoughts on how to proceed with a change to implement this?  I've implemented such a mechanism in a testing framework I worked on previously; however, each element was its own object with its own 'required' flag, whereas in SitePrism, individual elements are simply instantiated as a method call to Capybara and held in the containing section/page as a string in the mapped_items array.

IMO each element should be its own object and respond to `element#required?`, which defaults to true but can be set to false for stateful elements.

luke-hill:
Hi @tmertens You'll notice I've been bumping a few tickets recently and getting peoples input.

Can this and #49 be looked at and closed. I believe there is a PR tidying up `all_there?` slightly which will be merged soon.

luke-hill:
Closing as fixed - Released in site_prism `v2.11`

PR #129, opened 2015-09-24
Implement new Loadable behavior for pages and sections

tmertens:
@natritmeyer 

This PR adds a fully backwards compatible feature I call 'load validations' to SitePrism which we have been using internally for a while now with great success.  It provides a common abstraction for dealing with asynchronous behavior on pages, which can be a significant source of intermittent failures due to waiting for elements to be displayed.

This allows you to very cleanly abstract all of your waiting logic to a single place with a common pattern, and even allows for common validation logic to be shared between page if needed.  Additionally, it makes it very easy and clean to encapsulate this behavior in your tests without spreading boilerplate code throughout them.

The PR has thorough documentation and tests, please take a look.  I have additionally run our entire acceptance test suite against a branch of our SitePrism fork containing these changes, with the implementation of this in our test suite having been removed.

I can provide a more thorough contrived example if needed, but the basic gist assuming you have some pages defined with load validations is this:

```ruby
it 'does something' do
  home_page.load do |page|
    # Passing a block to the load method loads the page as usual,
    # then implicitly runs load validations before executing the block.
    # Existing functionality of the `load` method is not affected.
    page.register_button.click
  end

  registration_page.when_loaded do |page|
    # The page instance is yielded into the `load` and `when_loaded` blocks.  This allows you to cleanly encapsulate
    # interactions with each page, which is especially nice when performing multiple interactions
    # with a single page.
    page.email.set 'foo@example.test'
    page.password.set 'hunter2'
    page.submit
  end
  
  # Here we call a helper method on a page object which uses when_loaded to allow this to be executed as a one-liner in the test:
  another_page_with_a_form.fill_and_submit

  yet_another_page.when_loaded do |page|
    # Here we call a series of helper methods and/or element interactions on the page object.
    # Some of these methods may wrap their implementation in `when_loaded` blocks so that they
    # can be used safely as either a one-liner in a test, or as part of a larger construct as it is here.
    # The load caching in the Loaded module makes these nested calls extremely efficient, as only
    # the outermost `when_loaded` call actually execute the load validations while inner calls use the
    # cached result of the outermost block.
    page.do_thing_1
    page.do_thing_2
    page.complete
  end
end
```

We have found this to be very useful because:
* It removes boilerplate code for page readiness checks such as `page.displayed?` and/or testing for element existence in tests and/or helper methods prior to interacting with a page.
* It avoids nested/duplicate page readiness checks in test and helper methods, which incur a performance hit performing multiple duplicate queries against the driver.
* It provides a single, common abstraction for handling and using load validations for every page object.
* It makes our tests very clean with obvious scoping for the interactions with each page

I additionally believe this is a sufficient solution for issues #48, #49, and #50, as it enables users to define how they want a page to be tested for readiness which may include checks of 0 or more elements as needed and may include recursive checks against sections which may additionally provide their own load validations if desired.

It is also much more efficient and useful than `all_there?`, which does not provide an implicit wait mechanism and is a sledgehammer where a scalpel is needed.  Checking for all elements when testing if a page is ready for interaction under normal circumstances incurs a significant negative performance penalty for all tests that use it, where testing for URL and usually one or two elements is usually sufficient.

Please let me know your thoughts and comments.  From a functional standpoint, this PR is fully backwards-compatible, fully tested, thoroughly documented and should be ready to merge pending your approval.

Cheers!

### DIFF
--- a/lib/site_prism/exceptions.rb
+++ b/lib/site_prism/exceptions.rb
@@ -7,4 +7,5 @@ module SitePrism
   class TimeOutWaitingForElementVisibility < StandardError; end
   class TimeOutWaitingForElementInvisibility < StandardError; end
   class UnsupportedBlock < StandardError; end
+  NotLoadedError = Class.new(StandardError)
 end
--- a/lib/site_prism/loadable.rb
+++ b/lib/site_prism/loadable.rb
@@ -1,0 +1,84 @@
+module SitePrism
+  module Loadable
+    module ClassMethods
+      # The list of load_validations.  They will be executed in the order they are defined.
+      #
+      # @return [Array]
+      def load_validations
+        if superclass.respond_to?(:load_validations)
+          superclass.load_validations + _load_validations
+        else
+          _load_validations
+        end
+      end
+
+      # Appends a load validation block to the page class.
+      #
+      # When `loaded?` is called, these blocks are instance_eval'd against the current page
+      # instance.  This allows users to wait for specific events to occur on the page or certain elements
+      # to be loaded before performing any actions on the page.
+      #
+      # @param block [&block] A block which returns true if the page loaded successfully, or false if it did not.
+      def load_validation(&block)
+        _load_validations << block
+      end
+
+      def _load_validations
+        @_load_validations ||= []
+      end
+      private :_load_validations
+    end
+
+    def self.included(base)
+      base.extend(ClassMethods)
+    end
+
+    # In certain circumstances, we cache that the page has already been confirmed to be loaded so that actions which
+    # call `loaded?` a second time do not need to perform the load_validation queries against the page a second time.
+    attr_accessor :loaded, :load_error
+
+    # Executes the given block after the page is loaded.
+    #
+    # The loadable object instance is yielded into the block.
+    #
+    # @param block [&block] The block to be executed once the page has finished loading.
+    def when_loaded(&block)
+      previously_loaded = self.loaded # Get original loaded value, in case we are nested inside another when_loaded block.
+      raise ArgumentError unless block_given?
+
+      # Within the block, cache loaded? to optimize performance.
+      unless self.loaded = loaded?
+        message = "Failed to load because: #{load_error || 'no reason given'}"
+        raise ::SitePrism::NotLoadedError.new(message)
+      end
+
+      yield self
+    ensure
+      self.loaded = previously_loaded
+    end
+
+    # Check if the page is loaded.
+    #
+    # On failure, if an error was reported by a failing validation, it will be available via the `load_error` accessor.
+    #
+    # @return [Boolean] True if the page loaded successfully; otherwise false.
+    def loaded?
+      self.load_error = nil
+
+      return true if loaded
+
+      load_validations_pass?
+    end
+
+    # If any load validations from page subclasses returns false, immediately return false.
+    def load_validations_pass?
+      self.class.load_validations.all? do |validation|
+        passed, message = instance_eval(&validation)
+
+        self.load_error = message if message && !passed
+        passed
+      end
+    end
+    private :load_validations_pass?
+  end
+end
--- a/lib/site_prism/loadable.rb
+++ b/lib/site_prism/loadable.rb
@@ -43,16 +43,16 @@ module SitePrism
     #
     # @param block [&block] The block to be executed once the page has finished loading.
     def when_loaded(&block)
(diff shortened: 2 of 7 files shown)


almostwhitehat:
Thank you so much for this! This is something I've always implemented in a wrapper around loading pages - awesome to see it implemented this well right into site prism!

tmertens:
rubocop issues have been fixed.

tmertens:
@gerryster we chatted about this at windy city rails.

lib/site_prism/exceptions.rb
```diff
@@ -7,4 +7,5 @@ class TimeoutException < StandardError; end
   class TimeOutWaitingForElementVisibility < StandardError; end
   class TimeOutWaitingForElementInvisibility < StandardError; end
   class UnsupportedBlock < StandardError; end
+  NotLoadedError = Class.new(StandardError)
```
gerryster: I would suggest declaring this class as above for consistency.

tmertens: I would be more inclined to refactor the other error definitions since this way is cleaner :D if @natritmeyer agrees

natritmeyer:
Any PR that references James Bond in the tests is worthy of serious consideration :)

JanStevens:
Can this be merged released? Totally struggeling with bootstrap modals and waiting all over the place and not really getting anywhere :D

natritmeyer:
Just pushed the latest version

tgaff:
Ooooh this looks interesting @tmertens !  I'm going try to make time to read through the whole thing too in the next week or so!

tmertens:
@natritmeyer Any update on this?

natritmeyer:
Been playing around with it, looks great :) I've got a couple of friends testing it out and I'm waiting on feedback from them.

Anyone else had a play around with this? I'd be interested in what others think :)

Also, my wife was due to give birth yesterday so I'm kinda distracted ;)

tgaff:
@natritmeyer If it's happened already, congratulations! :smile: 

I'll admit I haven't looked **in-depth** yet but it's now officially on my TODO.  
(Shamelessly) if after the birth you somehow find time to merge/review mine as well I'd be pleased.  Although I suspect I'll have to redraw it after this.

natritmeyer:
@tgaff thanks, the boy has arrived safe and sound :)

@tmertens thanks for the reminder :) I've been spending the last week with my wife in the High Dependency Unit after she suffered some fairly major complications after an emergency c-section that went wrong. As excuses go, it's a pretty good one :)

I should be home in 2-3 days; I'll get to the rest starting from then :)

This comment has used up my daily emoticon quote

tmertens:
@natritmeyer That's why i didn't ask, I figured you were busy :)  I hope your wife is doing better now, and congratulations!

tgaff:
@tmertens Finally got a chance and ran into no issues; sorry for the continued delay.  I see this has been merged :+1:

lib/site_prism/loadable.rb
```diff
@@ -0,0 +1,84 @@
...
+    #
+    # The loadable object instance is yielded into the block.
+    #
+    # @param block [&block] The block to be executed once the page has finished loading.
+    def when_loaded(&block)
+      previously_loaded = loaded # Get original loaded value, in case we are nested inside another when_loaded block.
+      fail(ArgumentError, 'A block was expected, but none received.') unless block_given?
+
+      # Within the block, cache loaded? to optimize performance.
+      unless self.loaded = loaded?
```
natritmeyer: I'm getting the following rubocop warning:


lib/site_prism/loadable.rb:50:26: W: Assignment in condition - you probably meant to use ==.
      unless self.loaded = loaded?

tmertens: No, the behavior is correct, it is caching the result of `loaded?` but rubocop has a point.  This assignment could be moved either to the line before this expression, or to the end of the `loaded?` method:

```ruby
  self.loaded = loaded?
  unless self.loaded  # I don't recall if there was a specific reason for using explicit calls to `self` in this module, it is probably unnecessary.
    # ...
```

OR

```ruby
    def loaded?
      self.load_error = nil

      return true if loaded

      self.loaded = load_validations_pass?
      self.loaded
    end
```

tmertens: you could argue for renaming `loaded` as well to `loaded_cache` or similar to make it more clear that it is a cache of the load validation.

PR #130, opened 2015-09-24
Support older versions of Capybara with default_wait_time.

### LINKED ISSUE #128: Capybara.default_max_wait_time is not backwards compatible
I see there have been some recent changes to use `Capybara.default_max_wait_time` in place of `Capybara.default_wait_time`, but this change is backwards incompatible with older versions of capybara.

I think ideally all references to this should be abstracted to one place, such as `SitePrism::Waiter. default_wait_time` with a more forgiving method definition to handle both the new and deprecated Capybara configuration syntaxes in a backwards-compatible way.

tmertens:
This fixes #128.  It also updates the tests for `SitePrism::Waiter` to reflect the new behavior and makes other general improvements to the tests for that class.

This PR was verified against our acceptance test suite which currently uses Capybara 2.4.4 and fails when run against the SitePrism master branch.

### DIFF
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -99,7 +99,7 @@ module SitePrism
       method_name = "has_#{element_name}?"
       create_helper_method method_name, *find_args do
         define_method method_name do |*runtime_args|
-          wait_time = SitePrism.use_implicit_waits ? Capybara.default_max_wait_time : 0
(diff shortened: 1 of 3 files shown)


tmertens:
Fixed up the rubocop warnings from the original commit, build is passing now.

natritmeyer:
Thanks :)

PR #135, opened 2015-11-26
fix typo in README

cantonic:
I remember this confused me the first time I read it since I was thinking that the `NOT` in the comment refers to the syntax of `not_to` so I was thinking "what? of course it is not `not_to_`" ;)


natritmeyer:
Thanks :)

PR #137, opened 2015-12-12
Remove Ruby Warnings

murphyslaw:
When interacting with a `SitePrism::Page`, the warning `lib/site_prism/page.rb:8: warning: instance variable @page not initialized` is displayed, because the `@page` variable is used before it is assigned a value to. This fix initializes the variable in the `initialize` method. A different approach would have been to use `defined?(@page) ? @page : Capybara.current_session` in the `page` method, but I find the initializer to be more self documenting.


tmertens:
I agree an initializer is the right approach. Can you add a test case for this? Clearly there is a unit test missing here if it wasn't caught before

murphyslaw:
Please read my commit messages, as I tried to explain the changes and added tests in them.

lib/site_prism/page.rb
```diff
@@ -72,11 +75,11 @@ def self.set_url_matcher(page_url_matcher)
     end
 
     def self.url
-      @url
+      @url ||= nil
     end
 
     def self.url_matcher
-      @url_matcher || url
+      @url_matcher ||= nil || url
```
tmertens: First, the nil here will never be used since it will always hit the `||` in this case.

Second, I think it would be better to just initialize `@url_matcher` to `nil` since it is possible if the url_matcher is accessed before the url is set, then this could end up getting memoized as nil and then will never fall back to the `url`.

murphyslaw: As this is a class method, do you think we should do the following?

```
@url_matcher = nil
def self.url_matcher
  @url_matcher || url
end
```

luke-hill:
Hi @murphyslaw 

Apologies first of all for this being quite outdated. I was wondering where you feel this code is at at the present moment in time. Do you feel as though this is something you'd be willing to revisit / fix up and get ready to be merged into the codebase.

If you no longer feel this is relevant then could you close the PR

Thanks

Luke

PR #139, opened 2015-12-29
Wait for element stale

PikeIt:
Adding ability to wait for the element that takes a while to disappear

### DIFF
--- a/features/page_element_interaction.feature
+++ b/features/page_element_interaction.feature
@@ -35,6 +35,12 @@ Feature: Page element interaction
     Then when I wait for the element that takes a while to appear
     Then I successfully wait for it to appear
 
+  Scenario: Wait for element disappearing
+    When I navigate to the home page
+    Then I successfully see element that takes a while to disappear
+    Then when I wait for the element that takes a while to disappear
+    Then I successfully wait for it to disappear
+
   Scenario: Wait specific amount of time for element to appear
     When I navigate to the home page
     And I wait for a specifically short amount of time for an element to appear
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -50,6 +50,7 @@ module SitePrism
       create_existence_checker iframe_name, element_selector
       create_nonexistence_checker iframe_name, element_selector
       create_waiter iframe_name, element_selector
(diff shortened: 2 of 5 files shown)


features/page_element_interaction.feature
```diff
@@ -35,6 +35,12 @@ Feature: Page element interaction
     Then when I wait for the element that takes a while to appear
     Then I successfully wait for it to appear
 
+  Scenario: Wait for element disappearing
+    When I navigate to the home page
+    Then I successfully see element that takes a while to disappear
+    Then when I wait for the element that takes a while to disappear
```
alvy54: Should this be a When instead?

lib/site_prism/element_container.rb
```diff
@@ -131,6 +133,18 @@ def create_waiter(element_name, *find_args)
       end
     end
 
+    def create_disappearance_waiter(element_name, *find_args)
+      method_name = "wait_for_#{element_name}_disappearing"
```
alvy54: Should we call this wait_for_#{element_name}_stale instead?

features/page_element_interaction.feature
PikeIt: Yep, it is a copy/paste typo

lib/site_prism/element_container.rb
PikeIt: @alvy54 I think wait_for_#{element_name}_disappears sounds clearer. What do you think?

tmertens: Why not just use the `wait_until_myelement_invisible` method?

natritmeyer:
As @tmertens wrote, why not just use the `wait_until_myelement_invisible` method?

lib/site_prism/element_container.rb
PikeIt: Because sometimes i got follow error: 
`Failure/Error: organization_page.wait_until_sharing_settings_popup_invisible 
     Selenium::WebDriver::Error::StaleElementReferenceError:
       stale element reference: element is not attached to the page document
         (Session info: chrome=58.0.3029.81)`

luke-hill:
This doesn't really enhance the framework, so will reject the PR in favour of custom selectors

PR #140, opened 2016-01-08
Add Table of Contents to README

ddzz:
Table of contents generated with [DocToc](https://github.com/thlorenz/doctoc).  This table of contents will make it much easier to navigate site_prism's large README.


natritmeyer:
That's why I like http://www.rubydoc.info/github/natritmeyer/site_prism - autogenerated TOC that doesn't need to be maintained :)

Is there a ruby doctoc equivalent that could generate the TOC in a rake task, or something like that? It would be a shame to add `npm` to the project's dependencies :/

PR #141, opened 2016-01-21
Update page.rb

shabesoglu:
Returning self reference when loading a page.

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -32,6 +32,7 @@ module SitePrism
       end
 
       when_loaded(&block) if block_given?
+      self
     end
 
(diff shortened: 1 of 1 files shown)


tmertens:
Test?  You should also specify the `@return` value in the YARD documentation for the `load` method.

natritmeyer:
Agree with @tmertens. A test and some documentation would be great :)

luke-hill:
This is one of a few PR's raised which I'm looking into as Stale.

Could you please re-evaluate what has been asked of you and also if you feel this is still relevant.

Otherwise can you provide an update. Specifically to the Documentation requested and test coverage

PR #142, opened 2016-01-26
add implicit waits to the #all_there? check if they're enabled

bhaibel:
`use_implicit_waits` is a lifesaver for my particular usage of SitePrism. Its implementation should be more consistent. :)


natritmeyer:
The point of `all_there?` is to check that a bunch of elements exist, not to wait for them to exist...

tmertens:
@bhaibel You may also want to check out the new `load_validations` feature, which is a much more fine instrument than the blunt check for `all_there?`

Issue #102, opened 2015-02-03
Including RSpec::Matchers in Page classes will mess up `elements` and `sections`


yijia-zhan:
I want to move some assertions  to inside the Page class, and for that I need to `include RSpec::Matchers`. However, doing so will turn `elements` and `sections` into `RSpec::Matchers::BuiltIn::All`. 

For example:

```ruby
class HomePage < SitePrism::Page
  sections :menu_items, MenuItemClass, "ul#menu li"

  def test_method
    binding.pry
  end
end
```
In the code above, if I'm at the `binding.pry`, `menu_items` will return an array of menu items as expected. 

However, if I included `RSpec::Matchers` as following:

```ruby
class HomePage < SitePrism::Page
  include RSpec::Matchers

  sections :menu_items, MenuItemClass, "ul#menu li"

  def test_method
    binding.pry
  end
end
```

And if I  try to call `menu_items` at the `binding.pry`, I will get an error at `lib/site_prism/element_container.rb:33`:
>NoMethodError: undefined method `collect' for #<RSpec::Matchers::BuiltIn::All:0x007fd462d29ae0>

So if included `RSpec::Matcher`, instead of returning an Array, `find_all(*find_args, *runtime_args)` in the `sections` method will return a `RSpec::Matchers::BuiltIn::All` which does respond_to `:collect`. 

```ruby
def sections(section_collection_name, *args, &block)
    section_class, find_args = extract_section_options args, &block
    build section_collection_name, *find_args do
      define_method section_collection_name do |*runtime_args|
        find_all(*find_args, *runtime_args).collect do |element|
          section_class.new self, element
        end
      end
    end
  end
```

abotalov:
Duplicate of https://github.com/natritmeyer/site_prism/issues/95

abotalov:
I believe to fix it `all` should be changed to `page.all`.

yijia-zhan:
Thanks for the help @abotalov! I tested it out and it did make the code work. Not sure if there are side-effects of this change though. I'll send a pull request.

Issue #95, opened 2014-12-01
elements is an RSpec matcher class, rather than a collection


swolfekabaminc:
using the 'elements' method to set a collection of elements on a page. Rather than getting an array of elements, I'm getting an RSpec matcher:

    class ManagePages < BasePage
        elements :edit_links, 'a.edit'
    end

    ...
    @manage_page = ManagePage.new
    @manage_page.edit_links
    #<RSpec::Matchers::BuiltIn::All:0x00000104253e88 @matcher="a.edit", @failed_objects={}>

that last should give me an array, not an RSpec matcher ... ?

natritmeyer:
@swolfekabaminc if you include library A into library B and both have the same method name, you're going to get clashes.

PR #148, opened 2016-02-16
fixed rspec 3 method name collision

tobithiel:
I ran into the same issues as people in #95 and #102. These changes fix the name collision by explicitly selecting the recipient of the method call.
I'm not an expert with the inner workings of this gem and Capybara, so please let me know if I missed something.


tobithiel:
Forget this for now ^^

PR #153, opened 2016-02-26
Allow to set default expansion for url

ineverov:


luke-hill:
Hi @ineverov You've raised a PR (Admittedly nearly 2 years ago now), with no justification / docs / use cases.

Could you perhaps update with where you feel this improves the gem / repo and what use cases it solves. Perhaps through the use of an SSCCE.

Can you give us some sort of update also as to any changes you feel you would need to make in order to pass this into the repo.

If you don't feel this is relevant any more, could you close the PR?

Thanks,

Luke

PR #154, opened 2016-02-26
Allow to set custom max default wait time

ineverov:


luke-hill:
This is one of a few PR's raised which I'm looking into, and just generally giving a "bump".

Could you provide an update if at all possible into what you want to do with this PR (I'm aware it's been left a while)

Thanks for your contribution.

Note: I understand the changes made in this PR. But I don't understand the use-case. In-line waiters and SP's auto-waiter shouldn't really be altered. Just configure them privately inside your repo

ineverov:
@luke-hill yes sorry about it, it's completely my fault providing PR without justification. I'll try to do it now.
`Capybara.default_max_wait_time` if used mainly for implicit wait for element to be displayed. On the other hand `SitePrism.default_wait_time` is used in `page.rb` in method `displayed?` which check now that URL is changed. `SitePrism.default_wait_time` can also be uesd for waiting for `sections` and `pages` to be `loaded?` without explicit timeout. Like, `SitePrism::Waiter.wait_until_true { some_page.loaded? }`.

Acutally, it might be a good idea to allow `Loadable` objects provide it's own internal wait time.
For example,
```ruby
class SomePage << SitePrism::Page
  set_load_max_wait_time 20
  load_validation do
    [has_some_section?, 'Section is not displayed']
  end
  section :some_section, 'locator' do
     set_load_max_wait_time 5
     load_validation do
        [has_no_spinner?, 'Section is still loading... =(']
     end
     element :spinner, 'spenner_locator'
  end
end

page = SomePage.new
page.load # vist & wait_until_true(page.load_max_wait_time) { page.loaded? } & when_loaded(&block)
page.some_section # init section object & wait_until_true(section_object.class.load_max_wait_time) { loaded? }
```
Doing so will incapsulate all waitings inside page object. Please, let me know your opinion. I really like this gem and would like to contribute to it. Thanks, @natritmeyer !

luke-hill:
Quite a few of the items you have mentioned have either...
a) Changed since this post
b) Are going to change in the 2.x site_prism versions in discussion (2.13 and 2.14)
c) Are known issues (See TODO.md) and another issue. Whereby we are looking to fix them up there.

PR #158, opened 2016-03-21
Accessing a section's full text, HTML tag fixes

ddzz:
This PR adds a `#text` method to the Section class. The method returns
all of the text within a section. Without this method, calling `#text` on a
section would return all of the text on the page.

This PR also addresses minor issues with closing tags in the test site.
HTML tags were appearing when `page.text` was called on the test site
due to several missing tags.


natritmeyer:
Thanks :)

PR #160, opened 2016-05-04
Ensure sections process Loadable Validations

krimsonkla:
Readme states loadable validations can be added to Sections however once added, they are never being processed.


tmertens:
@krimsonkla The load validations on sections execute when either `loaded?` or `when_loaded` is called on the section.  It is not intended for load validations to run automatically every time the section is referenced.  This could lead to unexpected problems as well as very bad performance when using a section outside of a `when_loaded` block because it will run the validations on every single reference to the section without the benefits of caching that `when_loaded` gives you when executed with block notation.

This becomes especially problematic if you use load validations on a `sections` collection.  SitePrism will find all of the matching sections and run load validations against every single one before finally returning the resulting collection.  This would be extremely bad for performance if you are querying against a large set of sections.  You should only execute the load validations as often as is absolutely required to make your test robust against failure, and minimize as much as possible any superfluous queries against the DOM.

I don't think that executing the validations as part of the initializer is a good idea, although it _might_ be worth executing them implicitly (via `when_loaded`) around the `Capybara.within(@root_element)`  at the end of the initializer _if and only if_ a block is given.

PR #166, opened 2016-08-12
Improve find_first method to use first method from capybara

jtlimo:
Today when we use `element` in a section the SitePrism use capybara `find` method and raise an error if found more than 1 element, like that:

```
Capybara::Ambiguous: Ambiguous match, found 2 elements matching css "p.foobar"
```

My patch change the `find_first` method to use the method [`first`](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Node/Finders#find-instance_method) from capybara instead of `find`

### DIFF
--- a/lib/site_prism/section.rb
+++ b/lib/site_prism/section.rb
@@ -42,7 +42,7 @@ module SitePrism
     private
 
     def find_first(*find_args)
-      root_element.find(*find_args)
+      root_element.first(*find_args)
     end
 
     def find_all(*find_args)


almostwhitehat:
Is there a benefit to this over using the multiple element mapper (elements) and using #first on the result?

ineverov:
@jtlimo You can define your element/section with `match` option to get the same result, like this
`element :foo, 'p.foo-bar', match: :prefer_exact `

Assuming the first element when we have several like this, is not a good idea, IMO

BrianHawley:
This loses the feature of that `Capybara::Ambiguous` exception. You can accomplish the same thing, without losing the feature when you need it, by using the `match: :first` option of `find`.

luke-hill:
Thankyou for your PR. However I will be rejecting it in its current form.

Capybara is in a feature "slush" phase (So to speak). No huge changes coming whilst they debate what Capybara 3 will look like. One of the big proposed changes to their API is to rework `find_first`. See here: https://github.com/teamcapybara/capybara/blob/master/History.md specifically the 3.0.0.dev cut.

Furthermore your change would be a breaking one and require a 3.0 cut of our codebase, which we're not really in a position to do at the moment (Recovering from a long stagnant period of nearly 2 years without a gem update).

Thankyou for your interest. My advice for wanting to do this though is to code this at a local level. And have your codebase either include this as a module or inherit from a variant of `SitePrism::Page` / `Section`

Thanks

Luke

Issue #35, opened 2013-06-25
Content should be found within a section


dnesteryuk:
Hello guys,

There is a problem with SitePrism. When we write something like:

```ruby
@products.list.should have_content('Cars') # list object is a section
```

it will make capybara to look for this text on an entire page. It means we cannot make sure our text is found within a correct element. In addition, it slows down our tests since Capybara has to parse an entire page to find the text we are looking for.

At first, I had wanted to create a pull request, because I have the solution (at least it is some working idea), but then I decided to discuss it. My solution:

```ruby
module SitePrism
  class Section
    alias_method :origin_should, :should

    def should(matcher, &block)
      meth_name = matcher.instance_variable_get(:@expected).to_s.sub("have_","has_")
      meth_name << '?'

      if self.respond_to?(meth_name)
        self.origin_should(matcher, &block)
      else
        self.root_element.should(matcher, &block)
      end
    end
  end
end
```

I suggest to redefine the `should` method for the `SitePrism::Section` class and check if our section has a method which should be used by matcher, we pass matcher to the original `should`, but if a section doesn't have a method which should be used by a matcher, we pass a matcher to the `should` method of an original element of Capybara. By using this way, a searched looked content will be searched within a correct section.

What do you think? (please, ignore ugly way for getting the expected method name)

PR #174, opened 2017-01-04
Find content within a section

ilyasgaraev:
This PR related to issue #35 

If we write something like:
```ruby
expect(page.some_section).to have_content('Some content')
```
it will make capybara to look for this text on an entire page. But I think, many developers expect that `#have_content` method will search only in section.

Example, based on `TestHomePage` class:

**HTML**:
```html
...
<p>
  <a href='search.htm'>Search Page</a>
</p>
<article class='people'>
  <h1>People</h1>
  <span class='person'>person 1</span>
  <span class='person'>person 2</span>
  <span class='person'>person 3</span>
  <span class='person'>person 4</span>
</article>
...
```

**Code**:
```ruby
class TestHomePage < SitePrism::Page
  set_url '/home.htm'
  set_url_matcher(/home\.htm$/)
  ...
  section :people, People, '.people'
  ...
end
```

**Spec**:
```ruby
home = TestHomePage.new
home.load

expect(home.people).to have_content('Search Page')
```

Before my changes this spec finished without any failures, but now any capybara matchers scoped by section.

### DIFF
--- a/lib/site_prism/section.rb
+++ b/lib/site_prism/section.rb
@@ -17,12 +17,13 @@ module SitePrism
       Capybara.within(@root_element) { yield(self) } if block_given?
     end
 
-    def visible?
-      root_element.visible?
+    # Capybara::DSL module "delegates" Capybara methods to the "page" method
+    def page
+      root_element || Capybara.current_session
     end
 
-    def text
-      root_element.text
+    def visible?
+      page.visible?
     end
 
     def execute_script(input)
@@ -45,10 +46,6 @@ module SitePrism
 
     def find_first(*find_args)
       root_element.find(*find_args)
-    end
-
-    def find_all(*find_args)
-      root_element.all(*find_args)
     end
 
     def element_exists?(*find_args)
--- a/spec/section_spec.rb
+++ b/spec/section_spec.rb
@@ -63,5 +63,29 @@ describe SitePrism::Section do
       expect(section).to respond_to(:execute_script)
       expect(section).to respond_to(:evaluate_script)
     end
+
+    it 'responds to #visible? method' do
+      expect(section).to respond_to(:visible?)
+    end
+
+    it 'responds to Capybara methods' do
+      expect(section).to respond_to(*Capybara::Session::DSL_METHODS)
+    end
+  end
+
+  describe 'page' do
+    subject(:section) { SitePrism::Section.new('parent', root_element).page }
+
+    let(:root_element) { 'root' }
+
+    it { is_expected.to eq('root') }
+
+    context 'when root element is nil' do
+      let(:root_element) { nil }
+
+      before { allow(Capybara).to receive(:current_session).and_return('current session') }
+
+      it { is_expected.to eq('current session') }
+    end
   end
 end


lib/site_prism/section.rb
```diff
@@ -15,12 +15,12 @@ def initialize(parent, root_element)
...
     end
 
-    def visible?
-      root_element.visible?
+    def page
+      @root_element || Capybara.current_session
     end
 
-    def text
-      root_element.text
```
ilyasgaraev: `#text` method removed, because `Capybara::DSL_METHODS` includes `:text`:

https://github.com/teamcapybara/capybara/blob/30d037a0050085be996084bbf6538c3571bedea4/lib/capybara/session.rb#L34
https://github.com/teamcapybara/capybara/blob/master/lib/capybara/dsl.rb#L50

spec/section_spec.rb
```diff
@@ -49,9 +49,10 @@ class PageWithSection < SitePrism::Page
 
     context 'second argument is not a class and no block given' do
       it 'should raise an ArgumentError' do
-        class Page < SitePrism::Page
-        end
-        expect { Page.section :incorrect_section, '.section' }.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
```
ilyasgaraev: Too long :)

Gemfile
```diff
@@ -7,8 +7,8 @@ gem 'rake'
 gem 'selenium-webdriver'
 
 group :development do
-  gem 'yard'
   gem 'redcarpet'
-  gem 'simplecov', require: false
   gem 'rubocop'
+  gem 'simplecov', require: false
+  gem 'yard'
```
ilyasgaraev: Rubocop

spec/loadable_spec.rb
```diff
@@ -127,13 +127,13 @@ def foo
...
-      validation_spy_1 = spy(valid?: true)
-      loadable.load_validation { validation_spy_1.valid? }
+      validation_spy1 = spy(valid?: true)
+      loadable.load_validation { validation_spy1.valid? }
       instance = loadable.new
       instance.loaded = true
 
       expect(instance).to be_loaded
-      expect(validation_spy_1).not_to have_received(:valid?)
+      expect(validation_spy1).not_to have_received(:valid?)
```
ilyasgaraev: Rubocop

spec/loadable_spec.rb
```diff
@@ -16,17 +16,17 @@ def foo
     describe '#load_validations' do
```
ilyasgaraev: Rubocop

luke-hill:
Thankyou for your contribution. This seems a simple enough PR to land.

Can you do the following please
Remove 2nd commit.
Pull latest master
Rebase first commit with latest master `git rebase master`.
Repush back up so we have 1 nice clean commit to review.

It probably won't get looked at for a few weeks, because I'm just starting to look at PR's now, but if you're still interested we can look to get this reviewed and in soon.

ilyasgaraev:
Hi @luke-hill! Thank you for your comment. PR is ready for review, but unfortunately, rubocop fails with the following error:
```
Error: obsolete parameter EnforcedStyle (for Style/Encoding) found in .rubocop.yml
Style/Encoding no longer supports styles. The "never" behavior is always assumed.
RuboCop failed!
```

luke-hill:
:point_up: Is an issue that is probably pertinent to gemfile restrictions being lax and a removed value. There will be a bunch of "fix PRs" going up in 2.9.1 / 2.10 soon.

Thanks for getting your PR ready though

lib/site_prism/section.rb
```diff
@@ -15,12 +15,12 @@ def initialize(parent, root_element)
       Capybara.within(@root_element) { yield(self) } if block_given?
     end
 
-    def visible?
-      root_element.visible?
+    def page
```
luke-hill: I think I've worked this out.

Is the current code calling `page.find(matcher)` and as such you've overridden page at the Section level to provide the more granular version (localised element or if that isn't found then revert back to page)

spec/section_spec.rb
```diff
@@ -85,5 +86,26 @@ class Page < SitePrism::Page
       expect(section).to respond_to :execute_script
       expect(section).to respond_to :evaluate_script
     end
+
+    it 'responds to Capybara::Node::Element methods' do
+      expect(section).to respond_to :visible?
+      expect(section).to respond_to :text
```
luke-hill: I appreciate this is different from other areas of the code-base, But I'm gonna start getting a style into this suite slowly. Can we separate these expectations

spec/section_spec.rb
```diff
@@ -85,5 +86,26 @@ class Page < SitePrism::Page
...
+    end
+  end
+
+  describe 'page' do
+    subject(:section) { SitePrism::Section.new('parent', root_element).page }
+
+    let(:root_element) { 'root' }
+
+    it { is_expected.to eq 'root' }
```
luke-hill: Can we wrap both equality matchers in `()` (Again aware of differences, will be a tidy up PR from myself)

lib/site_prism/section.rb
luke-hill: edit: I've looked into the capybara docs and it does seem as though in the example you provided the failure is correct. But looking into the Capybara docs further it seems as though `have_text` which is an equivalent alias to `have_content` were only designed to be scoped at a page level. And infact it could potentially be fixed in the Capybara code.

luke-hill:
@ilyasgaraev Can you comment on https://github.com/teamcapybara/capybara/issues/1970 with your reproducible case. I want to get a part of this fixed at Capybara's end if it's still reproducible.

lib/site_prism/section.rb
ilyasgaraev: Hi, @luke-hill! Sorry for the late reply.

> only designed to be scoped at a page level

Could you please clarify, why do you think so? If this is because of the comment `Capybara::SessionMatchers`, I think this is just typo. Simple example:

*HTML*:
```
<div>
  <p class="sometext">Some text</p>
  <span>ANOTHER text</span>
</div>
```

*Code*:
```
node = page.find(".sometext")
expect(node).to have_text("ANOTHER")
```

This spec fails with the following error:
```
RSpec::Expectations::ExpectationNotMetError: expected to find text "ANOTHER" in "Some text"
```

The methods in https://github.com/teamcapybara/capybara/blob/master/lib/capybara/rspec/matchers.rb file are just matchers for rspec. 

`Capybara::DSL` module "delegates" `DSL_METHODS` to the `page` method (as you can see here: https://github.com/teamcapybara/capybara/blob/master/lib/capybara/dsl.rb#L49). I think, in Capybara terminology, the `page` method is a root level element. Therefore, I have defined `page` method in the Section class and this method returns `root_element` for this section. Now all capybara methods delegated to the section's root element. If you think that it is not clear, we can remove `Capybara::DSL` module and just "delegate" `NODE_METHODS` to the root_element (https://github.com/teamcapybara/capybara/blob/dfaf958ea7b9f02f31986f253ee10e6314853389/lib/capybara/session.rb#L41). DSL_METHODS includes NODE_METHODS, SESSION_METHODS, and MODAL_METHODS. And I think only NODE_METHODS make sense for our case. What do you think about this?

Thanks!

luke-hill: I posted an update below. And in capybara. They believe this "should" already happen.

I tested this a while ago on a version of capybara much older (I originally had this issue a year or two ago), Thomas is convinced this shouldn't happen and that all methods should be auto-scoped to whatever context you are calling them on.

Could you please post an SSCCE to the issue I mentioned where the current Test Case fails (i.e. it isn't being scoped correctly)

luke-hill:
Bump - Can you look into the request mentioned please. I've been reliably informed this "should" work. Can you let me know if it doesn't work as intended.

twalpole:
@luke-hill Looking further into this, the root cause is the inclusion of `Capybara::DSL` into `Section`.  Capybara::DSL contains methods that default to sending to Capybaras current session (which includes any scoping Capybara knows about from `within` etc) rather than the element specific versions of those methods if things like `Capybara::Node::Matchers` was included.  In Capybara v3 this could be solved by just implementing a `to_capybara_node` method in `Section` however as @ilyasgaraev mentions in v2 it's probably best to proxy all NODE_METHODS to root_element.

luke-hill:
Cheers @twalpole - So in its current guise this is ok and required. Think we need some heavy documentation around this. Appreciate the assistance here as much of this is new to me.

luke-hill:
Will hit the next RC (Release date yet TBC) - Thanks for your contribution.

nudded:
@luke-hill Doesn't this change break the usage of #within
For instance, if you want to add a method on a subclass of Section, that uses #within from Capybara, it will be sent to root_element! a Capybara::Node::Element does not understand this method. 

Am i missing something? or is this currently broken?

ilyasgaraev:
Hi @nudded! Could you please provide some code example?

luke-hill:
It uses `within` the `root_element` - which in this instance is the section.

If the `root_element` is undefined (Which I can't think of an example, but the coder introduced cover). We then call `super` (Or possibly will do soon)

nudded:
is the root_element the section? in our test-code this is actually a Capybara::Node::Element

PR #177, opened 2017-02-24
Lambda support inside page objects and sections

vgrigoruk:
Allows using lambdas for arguments of `set_url`,  `set_url_matcher`,  `element`,  `elements`,  `section`,  `sections` methods. This allows to defer the evaluation of argument value until the value is actually needed (now it is always done at the moment classes are loaded).

This allows to reuse the same page class if DOM / URL is different based on some conditions (while keeping the same interface).
Use case:
```
  element :some_element, -> { Someclass.some_property == some_val ? '#id1' : '#id2' }
end

# where Someclass.some_property is changed in a runtime
```

### DIFF
--- a/lib/site_prism.rb
+++ b/lib/site_prism.rb
@@ -4,6 +4,7 @@ require 'addressable/template'
 module SitePrism
   autoload :ElementContainer, 'site_prism/element_container'
   autoload :ElementChecker, 'site_prism/element_checker'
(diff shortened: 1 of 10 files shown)


vgrigoruk:
Any feedback @natritmeyer ?

luke-hill:
Refactor points requested. PR cannot be accepted in current form and as author seems absent closing PR

PR #178, opened 2017-02-24
Add a method for using within with an already existing section

cwitthaus:
We use a lot of ```sections``` in our site_prism code, and I was trying to use the ```within``` functionality to scope some expectations to a specific row in a table. Unfortunately, since the sections are initialized and placed into an array when using something like ```page.rows(text: "value").first``` it's not possible to add a block and execute code within the scope of the section. This change should leave the original behavior in place, and also add a method that allows users to scope a block for a ```section``` created by a ```sections``` declaration.

### DIFF
--- a/lib/site_prism/section.rb
+++ b/lib/site_prism/section.rb
@@ -9,10 +9,10 @@ module SitePrism
 
     attr_reader :root_element, :parent
 
-    def initialize(parent, root_element)
+    def initialize(parent, root_element, &block)
       @parent = parent
       @root_element = root_element
-      within { yield(self) } if block_given?
+      within(&block) if block_given?
     end
 
     def visible?
--- a/lib/site_prism/section.rb
+++ b/lib/site_prism/section.rb
@@ -12,7 +12,7 @@ module SitePrism
     def initialize(parent, root_element)
       @parent = parent
       @root_element = root_element
-      Capybara.within(@root_element) { yield(self) } if block_given?
+      within { yield(self) } if block_given?
     end
 
     def visible?
@@ -37,6 +37,10 @@ module SitePrism
         candidate_page = candidate_page.parent
       end
       candidate_page
+    end
+
+    def within
+      Capybara.within(@root_element) { yield(self) }
     end
 
     private
--- a/spec/section_spec.rb
+++ b/spec/section_spec.rb
@@ -70,7 +70,7 @@ describe SitePrism::Section do
 
   it 'passes a given block to #within' do
     block_variable = 5
-    expect_any_instance_of(SitePrism::Section).to receive(:within).and_yield
+    expect_any_instance_of(Capybara).to receive(:within).and_yield
     SitePrism::Section.new(a_page, 'div') { block_variable = 7 }
     expect(block_variable).to eq(7)
   end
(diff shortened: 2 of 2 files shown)


lib/site_prism/section.rb
```diff
@@ -12,7 +12,7 @@ class Section
     def initialize(parent, root_element)
       @parent = parent
       @root_element = root_element
-      Capybara.within(@root_element) { yield(self) } if block_given?
+      within { yield(self) } if block_given?
```
tmertens: This adds an additional/unnecessary extra proc around the block execution when it gets passed to `within`.  You should modify the initializer method here to specify the block in the parameter list, and pass the block to within as `&block`, which then will yield within the `within` method.

```ruby
def initialize(parent, root_element, &block)
  # ..
  within(&block) if block_given?
end
```

cwitthaus:
Are you working on cleaning up these rubocop errors in another branch, or just ignoring them for now?

spec/section_spec.rb
```diff
@@ -68,9 +68,11 @@ class Page < SitePrism::Page
     expect(SitePrism::Section).to respond_to :elements
   end
 
-  it 'passes a given block to Capybara.within' do
-    expect(Capybara).to receive(:within).with('div')
-    SitePrism::Section.new(a_page, 'div') { 1 + 1 }
+  it 'passes a given block to #within' do
+    block_variable = 5
+    expect_any_instance_of(SitePrism::Section).to receive(:within).and_yield
```
tmertens: use of `expect_any_instance` is discouraged. Also, this is stubbing the subject of the test.

The original expectation on `Capybara` should probably be kept but updated to use `and_yield` as you are doing here.

spec/section_spec.rb
```diff
@@ -85,5 +87,23 @@ class Page < SitePrism::Page
       expect(section).to respond_to :execute_script
       expect(section).to respond_to :evaluate_script
     end
+
+    it 'responds to within method' do
```
tmertens: This test is superfluous.

spec/section_spec.rb
```diff
@@ -85,5 +87,23 @@ class Page < SitePrism::Page
...
+      expect(section).to respond_to :within
+    end
+
+    describe '#within' do
+      it 'passes a given block to Capybara.within' do
+        expect(Capybara).to receive(:within).with('locator')
```
tmertens: This example is duplicated by the next one and should be deleted.

tmertens:
> Are you working on cleaning up these rubocop errors in another branch, or just ignoring them for now?

Likely they are from default settings in new versions of rubocop

luke-hill:
Hi there, thanks for your PR. Sorry it's taken a while to get around to looking at it. I've only recently come onboard here.

As it's a small PR. It's likely we can try get this in a 2.10 or 2.11 release. I'm not sure which one yet, as there's a lot of tidying up I want to do first.

If possible can you provide me with a small code example showing me where you want to use this going forwards. As I see it this is simply going to provide you with a was to pre-filter on initialization. But I'd rather see a working example if at all possible.

luke-hill:
Hi @cwitthaus If you're still floating around could you see if you can give us an example of the problem you're trying to solve with this PR. Just so we can triage and merge in ASAP.

Is there anything stopping you rescoping the sections once defined? Something along the lines of

```
def scoped_sections
  your_plural_sections.select { |section| section.boolean? }
end
```

PR #185, opened 2017-08-29
Fix #50 all_there? does not use optional implicit waits

### LINKED ISSUE #50: all_there? does not use optional implicit waits

RustyNail:
Fix #50 
- The `has_#{element}?` methods can now set the wait time with the `use_implicit_waits` option by #43.
- However, `all_there?` method can not use the `use_implicit_waits` option, because it is overwritten by `Capybara.using_wait_time(0)`.


luke-hill:
Thanks for your commit. Sorry about the delay hanging around getting it merged. I can see how it works simply enough.

Just so you've got a little idea of the roadmap. The plan is to release a 2.9.1 cut of the gem soon with almost no changes / doc changes and then another release (2.9.2 or 2.10), with a few stylistic tweaks.

I might try squeeze that into that release if at all possible. Thanks again

Luke

luke-hill:
Hi @RustyNail can you rebase this with latest master and then repush up the single commit.

This should do two things.

Ensure you're testing like for like (There has been 6 months since this last was raised)
Allow you to utilise the changes in other PR's including your Travis PR, which should allow the build to run cleanly on Travis and pass.

RustyNail:
Hi @luke-hill 

I rebased this with latest master.
And I confirmed two things.
  - I test to see if it is as expected.
  - The build of Travis CI is successful.

I conducted the following tests.


```
describe 'Verify using_wait_time is enabled' do
  example do
      set_url 'https://www.yahoo.co.jp/'
      element :searchbox, "#searchbox"
    end

    ## Transit to the TestPage after 10 seconds with Javascript if visit
    visit 'http://www.sky.sannet.ne.jp/masapine/java_swindow4.html'
    using_wait_time 30 do
      expect(TestPage.new).to be_all_there
    end
  end
end
```

luke-hill:
Merged. Thankyou for your commit.

Issue #183, opened 2017-06-05
Cannot use load validations without giving page a URL


TimKilcullen:
In the README under the 'Adding a URL' section it states that adding a URL is optional, but it appears that if you don't add a URL, you will always get an exception `SitePrism::NoUrlMatcherForPage` when attempting to use load validations. It looks like this is because every page gets a default `displayed?`  load validation. Should this default load validation even be added if a URL is not set for a page?

natritmeyer:
Good point. I'll take a look 👍

tmertens:
> Should this default load validation even be added if a URL is not set for a page?

The default validation should probably just return true if there is no URL set.

luke-hill:
Closing in favour of PR (Now reviewed)

PR #188, opened 2017-10-13
Enable to load html with block parameter

kei2100:
`SitePrism::Page#load` fails when passing html string and block parameter.

## Expected behavior
```ruby
site_prism $ bundle ex irb -r 'site_prism' -r 'capybara' -r 'capybara/dsl'
p = SitePrism::Page.new
p.load('<html><body>test</body></html>') { |p| p.text }
=> "test"
```

## Actual behavior
```ruby
site_prism $ bundle ex irb -r 'site_prism' -r 'capybara' -r 'capybara/dsl'
p = SitePrism::Page.new
p.load('<html><body>test</body></html>') { |p| p.text }

Traceback (most recent call last):
       16: from /Users/kei2100/.anyenv/envs/rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/cli.rb:424:in `exec'
       15: from /Users/kei2100/.anyenv/envs/rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:28:in `run'
       14: from /Users/kei2100/.anyenv/envs/rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:75:in `kernel_load'
       13: from /Users/kei2100/.anyenv/envs/rbenv/versions/2.5.0/lib/ruby/gems/2.5.0/gems/bundler-1.16.1/lib/bundler/cli/exec.rb:75:in `load'
       12: from /Users/kei2100/.anyenv/envs/rbenv/versions/2.5.0/bin/irb:11:in `<top (required)>'
       11: from (irb):7
       10: from lib/site_prism/page.rb:36:in `load'
        9: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:50:in `when_loaded'
        8: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:71:in `loaded?'
        7: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:76:in `load_validations_pass?'
        6: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:76:in `all?'
        5: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:76:in `each'
        4: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:77:in `block in load_validations_pass?'
        3: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/loadable.rb:77:in `instance_eval'
        2: from /Users/kei2100/repos/github.com/kei2100/site_prism/lib/site_prism/page.rb:11:in `block in <class:Page>'
        1: from lib/site_prism/page.rb:42:in `displayed?'
SitePrism::NoUrlMatcherForPage (SitePrism::NoUrlMatcherForPage)
```

## Environment
- OSX 10.12
- site_prism ver 2.9
- capybara ver 2.18.0


luke-hill:
Hi there. Thanks for this PR. Can you provide a quick SSCCE page where perhaps this would fail with the existing codebase and how your PR now fixes this issue

Thanks

Luke

kei2100:
Hi @luke-hill 
Thank you for your comment.
I updated the description of this PR.

luke-hill:
This "should" fix https://github.com/natritmeyer/site_prism/issues/183 - Not reviewed this yet though. Will do soon

spec/page_spec.rb
```diff
@@ -107,6 +107,13 @@ def foo?
         page.load { |p| p.foo? && true }
       end
 
+      it 'should allow to load html and yields itself' do
+        class Page < SitePrism::Page; end
+        page = Page.new
+        expect(page).to receive(:foo?)
```
luke-hill: Can we maybe remove this line as it's covered in other expectations - leave it as blank to leave a return link before rspec

spec/page_spec.rb
```diff
@@ -107,6 +107,13 @@ def foo?
...
+        expect { page.load('<html/>') { |p| p.foo? && true } }.to_not raise_error
```
luke-hill: I'd prefer you just passing in the html - remove the block argument as that's not under test here.

Also could we verify the yielded value is the string argument itself.

luke-hill:
@tgaff and @TheMetalCode to review also ideally (Only a small ticket) - Get into good habits.

luke-hill:
Also aside from points already raised could you
1 - pull latest master
2 - rebase with latest master
3 - force push changes back up (So we don't get any excess baggage)

This will fix your Travis issues (Hopefully)

spec/page_spec.rb
tgaff: @luke-hill maybe I mis-understood you.  Based on the test name I think testing that the block is executed is what s/he intended.  
That being said I don't think that this test actually proves what it intends to.  If the block is completely ignored the test will still pass.

I've never used this feature, myself, but based on the original issue it seems it could be something like:
`expect { page.load('<html>hi</html>') { |p| p.text } }.to eq 'hi'`

luke-hill: Good point. I missed that. But yeh we definitely need a yielded arg

kei2100:
Thank you for your review.
I improved the `page_spec.rb`

luke-hill:
Not sure you were confident with cleanup tasks so I made https://github.com/natritmeyer/site_prism/pull/209 and amended some documentation

PR #193, opened 2018-02-16
Dev update 1 of 3

luke-hill:
One of the first which will land into the 2.9.2+ release candidate.

Primarily a dev based update this features no changes aside from patch based changes
- Rework all text files into Markdown structure, amending styles mainly where appropriate / spotted inconsistencies
- Amended / Added a small amount of data to the text files to keep community updated with next short term plans
- Added a couple of local gitignore files which will aid development and prevent extra items being un-necessarily pushed up
- Add a physical cap to Cucumber and Selenium Integration test runs (**Note: This is only meant as an interim stop-gap measure, and should be upped when appropriate**)
- Fix rubocop not executing
- Fix up 90%+ of failing cops (Just `MagicComment` and `SpaceAfterMagicComment`) _Note this is what gives the PR the 50 files changed moniker_
:point_up: Is coinciding with the Suite having inconsistent dependencies RE Gems / Ruby Versions. This will be tidied up going forwards. With a view in a release down the line to set the `.ruby-version` Higher, running travis back on multiple rubies and bumping the supported version/s


LICENSE.md
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2014, Nathaniel Ritmeyer
+Copyright (c) 2011-2016, Nathaniel Ritmeyer
```
tgaff: 2018?

Gemfile
```diff
@@ -1,10 +1,12 @@
+# frozen_string_literal: true
+
 source 'https://rubygems.org'
 
 gemspec
 
-gem 'cucumber'
+gem 'cucumber', '~> 2.4.0'
```
tgaff: Maybe these should actually be moved to the gemspec file?  I think the best practice currently is to have them in the gemspec - although that does seem to have been a point of contention for some time.

Admittedly 8 years old:
http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

luke-hill: I'm going to push a lot of this work into additional TODO's for several reasons
1) It's the best thing to try to tackle items in isolation
2) Some of these have been long-standing. And I'm still keen on only changing the bits I'm fully comfortable with
3) Whilst travis is being reworked, I'm a bit apprehensive on changing anything structural

Hope that's ok

PR #196, opened 2018-02-19
Added base contributing / issue templates

luke-hill:
To land in either 2.10 or 2.11 next month

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 # PR template for SitePrism framework
-Thanks for offering to help and raise a PR for the SitePrism framework, we have a few simple requests to make managing the repo easier going forwards.
+Thanks for offering to help by contributing a PR to site_prism.
+
+Please help us understand your change by including the following information in your PR.
 
 ## Type of PR
 - Feature / Bugfix / Refactor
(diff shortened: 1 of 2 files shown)


luke-hill:
@tgaff @TheMetalCode @RustedNail As you've all recently expressed desire to help out here Figured I'd ask you guys for your thoughts.

@tmertens @abotalov Not sure if you guys are still around? But hopefully this is going to get back off the ground if you're interested in sticking around?

tgaff:
Good idea.  
I have a few grammatical suggestions.  I think we probably speak different English variants, so please don't be surprised or offended if what sounds slightly off to me seems correct to you.  


> "Thanks for offering to help and raise a PR for the SitePrism framework, we have a few simple requests to make managing the repo easier going forwards."

I mis-read this the first couple of times; it might be a bit long.  I read "we have a few simple requests to make." + sentence fragment "managing the repo easier going forwards"
Perhaps this could be broken into a couple of sentences.  "Going forwards" is also somewhat superfluous.   
Possible alternative: "Thanks for offering to help by contributing a PR to site_prism.  Please help us understand your change by including the following information in your PR."

> http://sscce.org/

You might want to make `http://sscce.org/` into a link.

> Before raising an issue with us please ask yourself whether you have all the information to hand to advise us.

"to hand" and "at hand" seem to differ between AE and BE.  Perhaps this can be replaced with "available".


> For us to try help you with your issue. 

This is a fragment (and is missing a "to").  
Maybe: "For us to try to help you with your issue, we require certain information."  Or "We require some information in order to try to help you with your issue."


> Specificially `site_prism` and `capybara`)  

Would adding `rspec` to this be a good idea?  Also, I think there may be a typo in the spelling of "specifically".


> What behaviour did you observe you expecting?

What behaviour did you observe?


> If you can't provide a link to the page, consider creating a reproducible page on https://jsfiddle.net/

How about: 
If you can't provide a link to a page that the problem can be reproduced on, please consider creating a reproducible page on https://jsfiddle.net/

PR #199, opened 2018-02-20
Make Travis Green

luke-hill:
Removed rubocop compliance and sign-off whilst we get suite back under control

PR #201, opened 2018-02-21
Fix rubocop

RustyNail:
Hi @luke-hill 
Thank you for answering [my question](https://github.com/natritmeyer/site_prism/pull/199#issuecomment-367259959).
I solved the identification by rubocop.

#### 1. Fixed version of rubocop with '0.50.0
I fixed version of rubocop with '0.50.0', because the Ruby version specified in 'site_prism.gemspec' and the supported by latest rubocop did not match.
It caused the following points to be pointed out.

```
site_prism.gemspec:8:29: C: Gemspec/RequiredRubyVersion: required_ruby_version (2.0, declared in site_prism.gemspec) 
  and TargetRubyVersion (2.1, declared in .rubocop.yml) should be equal.
  s.required_ruby_version = '>= 2.0'
```

Although I thought of several countermeasures,I fixed it as fixed version by referring to the following issue.
Reference issue : https://github.com/bbatsov/rubocop/issues/5260#issuecomment-352151774

#### 2. Fixed identification by rubocop
- Improved with 'bundle exec rubocop -a'.
- Easy fix by manual work.
- Explicitly specify Ruby version in .rubocop.yml.

#### 3. Added setting in Rakefile
The setting was deleted by [this PullRequest](https://github.com/natritmeyer/site_prism/pull/199) but it was added because the rubocop test was restored.
If you prefer not to change, I revert the commit.

### DIFF
--- a/features/step_definitions/page_section_steps.rb
+++ b/features/step_definitions/page_section_steps.rb
@@ -113,7 +113,7 @@ end
 Then(/^I can see individual people in the people list$/) do
   expect(@test_site.home.people.individuals.size).to eq(4)
   expect(@test_site.home.people.individuals(count: 4).size).to eq(4)
-  expect(@test_site.home.people).to have_individuals count: 4
+  expect(@test_site.home.people).to have_individuals(count: 4)
 end
 
 Then(/^I can get access to a page through a section$/) do
--- a/features/step_definitions/page_section_steps.rb
+++ b/features/step_definitions/page_section_steps.rb
@@ -113,7 +113,7 @@ end
 Then(/^I can see individual people in the people list$/) do
   expect(@test_site.home.people.individuals.size).to eq(4)
   expect(@test_site.home.people.individuals(count: 4).size).to eq(4)
-  expect(@test_site.home.people).to have_individuals count:(4)
+  expect(@test_site.home.people).to have_individuals count: 4
 end
 
 Then(/^I can get access to a page through a section$/) do
--- a/spec/section_spec.rb
+++ b/spec/section_spec.rb
@@ -53,7 +53,9 @@ describe SitePrism::Page do
       it 'should raise an ArgumentError' do
         class Page < SitePrism::Page
         end
-        expect { Page.section :incorrect_section, '.section' }.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
+        expect do
+          Page.section :incorrect_section, '.section'
+        end.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
       end
     end
   end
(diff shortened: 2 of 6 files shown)


spec/section_spec.rb
```diff
@@ -53,7 +53,9 @@ class PageWithSection < SitePrism::Page
       it 'should raise an ArgumentError' do
         class Page < SitePrism::Page
         end
-        expect { Page.section :incorrect_section, '.section' }.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
+        expect do
```
luke-hill: Can we rewrite this spec as `{}` syntax but then do trailing `.` for the two args (Should really pass `with_message` as well as separate arg)

spec/section_spec.rb
```diff
@@ -53,7 +53,9 @@ class PageWithSection < SitePrism::Page
...
+          Page.section :incorrect_section, '.section'
```
luke-hill: `Page.section(:incorrect_section, '.section')`

features/step_definitions/page_section_steps.rb
```diff
@@ -113,7 +113,7 @@
 Then(/^I can see individual people in the people list$/) do
   expect(@test_site.home.people.individuals.size).to eq(4)
   expect(@test_site.home.people.individuals(count: 4).size).to eq(4)
-  expect(@test_site.home.people).to have_individuals count:(4)
+  expect(@test_site.home.people).to have_individuals count: 4
```
luke-hill: `have_individuals(count: 4)`

spec/section_spec.rb
RustyNail: I thought of several ways.
I selected a method using `{}` syntax.

[Method 1] Selected

```
context 'second argument is not a class and no block given' do
  subject(:section) { Page.section(:incorrect_section, '.section') }

    class Page < SitePrism::Page
    end
    expect { section }.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
  end
end
```

[Method 2]

```
context 'second argument is not a class and no block given' do
  subject { -> { Page.section(:incorrect_section, '.section') } }

    class Page < SitePrism::Page
    end
    is_expected.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
  end
end
```

RustyNail:
I corresponded to the pointed out matter.
please check them @luke-hill .

luke-hill:
@tgaff Any points to raise?

tgaff:
Thanks you for the thorough explanation and for fixing up rubocop @RustyNail !
This looks good to me.  Ruby 2.0 is a bit old, but if there is ever a need to update we can deal with that when the time comes.

luke-hill:
Just caught up on comments here. Think we'll start phasing out some of these down the line. I think for now I just want to keep the status quo (As it feels right not to have too many changing parts at once). I think I'll do the usual deprecate / remove protocol, but I'll add in an extra "unsupported" layer, when we remove a testing aspect, but allow runtime but without offering help.

PR #204, opened 2018-02-23
Travis Update - Take 2

luke-hill:
Unignore `.ruby-version` - We want people running dev on 2.0 forced (As we technically support this)
Add in latest 2.0 - 2.2 rubies. Update 2.3 ruby. (This now runs 4 sets of rakes as opposed to 1)
Update gecko from 0.18 to 0.19.1


luke-hill:
@RustyNail @tgaff can you let me know if you're happy with this.

Plan is for next RC (2.11)

tgaff:
@luke-hill looks great!

.ruby-version
```diff
@@ -0,0 +1 @@
+ruby-2.0
```
tgaff: Is this necessary?  I thought you recently removed this in a different PR.  

I think it will (annoyingly) lock us to developing in just ruby-2.0 version.  (Which I don't even have installed).  
I don't think this sets just a minimum.

Given that travis.yml already defines it's versions and the gemspec does so also, I think this is unnecessary.

luke-hill: Whilst we support ruby 2.0 We should be developing and testing against it.

I agree we need to migrate upwards and perhaps start deprecating or having unconfirmed support for these versions (Permitting the run-time of the gem at say ruby 2.0, but not guaranteeing support).

But at the moment we officially support ruby 2.0 - And I was asked not to make too wide sweeping changes yet.

Are you developing on MAC/Ubuntu (If so something like RVM hooks into this directly).

tgaff: Yes, I'm familiar with RVM, chruby and friends.  I get where you're coming from on wanting to prevent devs from accidentally doing something that might not work in 2.0, but that's pretty rare and I think just testing against 2.0 on travis is as likely to catch that as a dev would be.  
I don't know how common this is, but Capybara at least has this file .gitignore'd.  Either way, it's at worst an inconvenience, so feel free to keep it if you want.

Everything else here is great.  Glad Travis will be covering multiple rubies.

PR #205, opened 2018-02-23
Gemfile -> Gemspec migration

luke-hill:
Added low (2015 end), …versions where none specified.

This will conflict with #201 so will be merged in after it (Possibly in `2.11`, if not in `2.12`) - Dependent on other feature / refactor work and time between revisions (Want to keep it below a few weeks in between version bumps)

Bumps to the variety of dependencies which are either fixed or well below their best to come in future releases

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,15 +3,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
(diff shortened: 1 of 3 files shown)


luke-hill:
@RustyNail @tgaff @TheMetalCode quick review ticket.

PR #207, opened 2018-02-23
Refactor element checks sans metaprogramming

Kache:
clearly disambiguates the element name as the method argument

also prevents possible element name collision, consider:

    class MyPage < SitePrism::Page
      element :foo, '#foo'
      element :no_foo, '#no_foo'
    end

    page.has_no_foo? # ambiguous collision

previously metaprogrammed methods:

    page.has_foo?
    page.has_no_foo?
    page.wait_for_foo
    page.wait_until_foo_visible
    page.wait_until_foo_invisible

now:

    page.has?(:foo)
    page.has_no?(:foo)
    page.wait_for(:foo)
    page.wait_until_visible(:foo)
    page.wait_until_invisible(:foo)

### DIFF
--- a/features/step_definitions/iframe_steps.rb
+++ b/features/step_definitions/iframe_steps.rb
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 
 Then(/^I can locate the iframe by id$/) do
-  @test_site.home.wait_for_my_iframe
+  @test_site.home.wait_for(:my_iframe)
 
-  expect(@test_site.home).to have_my_iframe
+  expect(@test_site.home.has?(:my_iframe)).to be true
 end
 
 Then(/^I can locate the iframe by index$/) do
-  @test_site.home.wait_for_index_iframe
+  @test_site.home.wait_for(:index_iframe)
 
-  expect(@test_site.home).to have_index_iframe
+  expect(@test_site.home.has?(:index_iframe)).to be true
 end
 
 Then(/^I can see elements in an iframe$/) do
@@ -20,6 +20,6 @@ end
 
 Then(/^I can see elements in an iframe with capybara query options$/) do
   @test_site.home.my_iframe do |f|
-    expect(f).to have_some_text(text: 'Some text in an iframe')
+    expect(f.has?(:some_text, text: 'Some text in an iframe')).to be true
   end
 end
--- a/features/step_definitions/page_element_interaction_steps.rb
+++ b/features/step_definitions/page_element_interaction_steps.rb
@@ -9,15 +9,15 @@ Then(/^the page has no title$/) do
 end
 
 Then(/^the page does not have element$/) do
-  expect(@test_site.home.has_no?(:nonexistent_element)).to be true
+  expect(@test_site.home.no?(:nonexistent_element)).to be true
 
-  expect { @test_site.home.has_no?(:nonexistent_element) }.not_to raise_error(SitePrism::NoSelectorForElement)
+  expect { @test_site.home.no?(:nonexistent_element) }.not_to raise_error(SitePrism::NoSelectorForElement)
 end
 
 Then(/^the page does not have elements$/) do
-  expect(@test_site.home.has_no?(:nonexistent_elements)).to be true
+  expect(@test_site.home.no?(:nonexistent_elements)).to be true
 
-  expect { @test_site.home.has_no?(:nonexistent_elements) }.not_to raise_error(SitePrism::NoSelectorForElement)
+  expect { @test_site.home.no?(:nonexistent_elements) }.not_to raise_error(SitePrism::NoSelectorForElement)
 end
 
 Then(/^I can see the welcome header$/) do
@@ -33,7 +33,7 @@ Then(/^I can see the welcome header with capybara query options$/) do
 end
 
 Then(/^the welcome header is not matched with invalid text$/) do
-  expect(@test_site.home.has_no?(:welcome_header, text: "This Doesn't Match!")).to be true
+  expect(@test_site.home.no?(:welcome_header, text: "This Doesn't Match!")).to be true
 end
 
 Then(/^I can see the welcome message$/) do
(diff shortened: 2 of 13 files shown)


Kache:
I understand that this is a notable user-visible api change probably more suitable [in a major version](https://github.com/natritmeyer/site_prism/pull/207)

Kache:
Additional thoughts:

I thought about making the validators accept objects directly:

```ruby
page.has?(foo)
page.has_no?(foo)
page.wait_for(foo)
page.wait_until_visible(foo)
page.wait_until_invisible(foo)
```

Although I'd like to get some feedback before I commit more effort into it.

I would also like to suggest removing `has_no?` altogether since `!page.has?` should be sufficient. Note that I removed `SitePrism::Section#element_does_not_exist?` in this PR, which may have had a bug:

```ruby
    def element_does_not_exist?(*find_args)
      # should this be root_element.nil? || root_element.has_no_selector?(*find_args)
      root_element.has_no_selector?(*find_args) unless root_element.nil?
    end
```

luke-hill:
Points to consider (AS you've made a lot)

- Recently I have taken over the repo and would rather not make sweeping changes.
- You've included a lot of work (Whilst the +/- might not suggest it)
- You've proposed bugfixes / issues / rewriting of large portions of the API

Suggestions:

The bugfix issue you mentioned is almost a non-priority one (Returning `nil` opposed to `false` however it does merit a fix.
Each individual use-case can be triaged / fixed independently
The `has_no` name clash is a known issue and one most people would probably spot inside the API. Again is it an issue, yes. But I'd argue not a priority one.

If you're passionate about these rewrites of the API Could I suggest bucketing them into smaller PR's to review, where we can see features and code changes in isolation. It makes reviewing for the people easier (Remember that this repo was essentially managed by one person prior to me coming onboard, who wrote essentially the whole API)

Kache:
**Bugfix**

I considered the bugfix a higher priority because IT IS NOT a `nil` for `false`:

```ruby
# implementation of element_does_not_exist?
root_element.has_no_selector?(*find_args) unless root_element.nil?

# equivalent
if root_element.nil?
  nil # should this not return true? the "element does NOT exist"!
else
  root_element.has_no_selector?(*find_args)
end
```

A lot of effort could be made here to fix the bug while writing the test to enforce the new code. I just thought all that effort could be saved by simply deprecating the method.

**Independent Use-Cases**

Would you partition the use-cases as such:  
 * `has?`/`has_no?` (simultaneously deprecating `has_no?`)
 * `wait_for`
 * `wait_until_visible`/`wait_until_invisible`

If so, then yes, I can break up this PR into de-metaprogramming those, one at a time.

**Object-based API variant**

`page.has?(:foo)` vs `page.has?(foo)` -- Your thoughts?

luke-hill:
For now I've not spent a lot of time diving into your efforts / code. And would welcome PR's. However as it seems as though you've given this a lot of thought it's also worth me being prudent on where I stand on such work.

Item 1 - The bugfix. As a standalone PR any fix for this would be merged in the active branch and hit the next RC
Item 2 - Independent Use Cases - Your partitions are very good yes, I'm happy with those.
Item 3 - Object based variant (Favour symbols until we hit Ruby 3), smaller changes

Now for the big crux. Any API style changes wouldn't hit Siteprism 3. And as such there is no plans for us even to introduce a feature branch for this for several months (Honesty I think is the best policy).

In considering the work you're proposing I have some options that could be considered

a) Introducing these elements side by side duplicating the codebase, It gives people some options to consider - However It would also duplicate the doc requirements (Which we enforce)
b) Phase 1 out for 1 out. Which as you've mentioned makes the API inconsistent. Again not sure.

In summation, any of the API changes wouldn't get looked at for a long time. If you're passionate about contributing, we have an extensive Task list growing daily in our https://github.com/natritmeyer/site_prism/blob/master/TODO.md and would welcome any contributions against all of those.

Also as a brief history, Siteprism was inactive for the majority of 2016/2017, so one of the next main things we're looking to do is just re-establish ourselves.

Thankyou for your support. It's likely I won't be able to look at this thread again this week. So sorry if it seems I am distant.

luke-hill:
Also just to add some clarity RE the bugfix

```
# implementation of element_does_not_exist?
root_element.has_no_selector?(*find_args) unless root_element.nil?

# equivalent
if root_element.nil?
  nil # shouldn't this return true? the "element does NOT exist" under this root element!
else
  root_element.has_no_selector?(*find_args)
end
```

It is correct to return `false`, it returns `nil` which is slightly different, and a minor bug. Because the section does not have a root element, the section likely doesn't exist.

If a section doesn't exist. There is no valid test for ascertaining whether an element inside this section exists. As such the behaviour (false/nil aside), is correct.

Kache:
I think that decision is incorrect in multiple ways, and I think another opinion should be brought in on this.

As a user, I'm interested in asking the natural language question, "does child under its parent exist?" in the same way a CSS selector would: `#parent #child`. This question can be represented in the following truth table:

| parent exists | child exists | `SitePrism#element_exists?`<br/>"child under its parent exists?" | `!SitePrism#element_exists?`<br/>"child under its parent missing?" |
| :-: | :-: | :-: | :-: |
| T | T | T | F |
| T | F | F | T |
| F | T | F | T |
| F | F | F | T |

I am _not_ asking:

| parent exists | child exists | `SitePrism#element_does_not_exist?`<br/>"parent exists and child missing?"<br/>"if parent exists ➞ child missing" | `!SitePrism#element_does_not_exist?`<br/>"parent exists and child exists?"<br/>"if parent exists ➞ child exists"
| :-: | :-: | :-: | :-: |
| T | T | F | T |
| T | F | T | F |
| F | T | F | T |
| F | F | F | T |

I wouldn't use SitePrism to handle [three-valued logic the way RDBs do](https://en.wikipedia.org/wiki/Three-valued_logic#Application_in_SQL) nor would I expect it to implement [conditional statement logic](https://en.wikipedia.org/wiki/Conditional_sentence). Even if I did, SitePrism is bewilderingly implementing one from the top table and using one from the bottom table as its inverse.

As-is, I find this situation highly demotivating.

luke-hill:
Your top truth table is incorrect, as what you are evaluating is the presence of the following pseudo css element ('.parent .child') -> Where `parent` and `child` can be evaluated as different css selectors than the ones given.

This is not what siteprism is evaluating. At each evaluation site-prism returns an object. The way siteprism would build a relational object like this is not doing `page.find(.parent .child)` it actually will do the following.

`page.find(.parent).find(.child)` -> There is a subtle difference. Because at the first evaluation point we don't have an answer. The element is not found. So as mentioned, it cannot then perform the second evaluation.

As your top truth table mimicks a question which we don't currently ask (In our Section/Element structure), the argument isn't going to hold for this evaluation in a predicate.

Furthermore I feel as though we've now gone quite off-topic from the original PR of rebuilding parts of Site-prism. There is a reasonably large TODO list with plenty of rebuild / extension tasks that are able to be tackled should you wish. But for now I'm going to close this PR.

PR #208, opened 2018-02-25
Allow for expected_elements declaration to enhance all_there check

TheMetalCode:
This should be considered as an alternative approach to https://github.com/natritmeyer/site_prism/pull/98

Let's say the page I'm modeling with a SitePrism::Page class has at least one element that isn't always guaranteed to be present.

``` ruby
  element :name_field, '#name'
  element :err_msg, '.error'
end
```

If I ever wished to leverage that page's all_there? check, it would fail because the err_msg would not be present on initial load (hopefully). And yet, I would very much like to use the page's all_there? check.

This pull request adds the ability for any page class to declare `expected_elements` that specifies exactly which page elements should be considered for the `all_there?` check.

After some thought, I might personally prefer this approach over `excluded_elements` because, in my experience, it's less verbose to specify the elements that always should be present than to specify those that might not always be present. Furthermore, `expected_elements` does not need to include literally all elements that should be present - it can include only those element sufficient to signify that the page is ready to receive further automation actions. This is the approach I've taken when doing web automation in non-ruby contexts, and would like to propose a similar approach here.


lib/site_prism/element_container.rb
```diff
@@ -23,6 +23,10 @@ def elements(collection_name, *find_args)
     end
     alias collection elements
 
+    def expected_elements(*elements)
+      @expected_items = elements.map(&:to_s)
```
luke-hill: If these aren't mapped to strings and remain as symbols does anything happen?

luke-hill:
@tgaff Does this look good to you?

tgaff:
@luke-hill I won't have time to look at the code at least until tomorrow.  At first glance though I'm concerned that it might be a breaking change that will require current users using `all_there?` to refactor.  **Does it by default, default to the current `all_there?` behavior of checking every element?**

If not, and if I'm understanding this, now you have to specify a bunch of elements in order to use it, which will break existing code-bases.  In addition to that, having `all_there?` by default not really be "all there" strikes me as semantically confusing.  It's not what you would expect when reading just the test code.  (sometimes that's OK, but we should think hard about it)  It's one thing if it excludes one element of many, but a whole other thing if all_there? suddenly means 2 elements out of 20....

I had an idea for an alternate implementation that would allow defining groups of elements to check.  However, I also suspect this feature is getting far more attention than maybe it should.  Nat had stated that he was opposed to it even being in the code-base and I've never needed it.

TheMetalCode:
@tgaff No, it isn't a breaking change. `expected_elements` is optional, just as `excluded_elements` was on #98. We know this for two reasons:
 
First, look at how we gather the elements for the `all_there?` check:

```ruby
module SitePrism
  module ElementChecker
    def all_there?
      elements_to_check.all? { |element| send "has_#{element}?" }
    end

    private

    def elements_to_check
      elements = self.class.mapped_items # all defined elements on the page by default
      elements = elements.select { |el| self.class.expected_items.include?(el) } if self.class.expected_items # scope it to expected_elements only if that method is defined
      elements
    end
  end
end
```

Second, if it were a breaking change I would have had to change existing tests that cover `all_there?`, and I didn't. Build is green.

This is pretty much exactly the same logic as #98, we're just naming the elements we want to check instead of the ones we don't.

luke-hill:
I'm going to go ahead and merge this in as is. I understand specifically how it works. coupled with the `elements` third definition. As the second definition is explicitly ignored when there is no values in the DSL, so the third value basically "Does nothing"

lib/site_prism/element_container.rb
luke-hill: @TheMetalCode before I merge in can you just answer this question

TheMetalCode: @luke-hill Sorry, missed this earlier. The reason we want `to_s` here is that they end up getting evaluated in the `all_there?` check like so: `elements_to_check.all? { |element| send "has_#{element}?" }`. So if `element` wasn't already a string at that point, that `send` call wouldn't work.

luke-hill: For reference they "should". Interpolating symbols gets converted into strings and even running send on symbols should work. 

But hey I've merged this so it doesn't matter now.

tgaff:
@TheMetalCode Sorry, maybe I wasn't clear.  I had not had time to look at the code yet.  I was just asking because it's something I would want to be on the lookout for when I do read the code.

Can we talk about tidying up PRs and commit messages a little more in the future?  I think ideally they could be more succinct.  This PR has 27 commits, a number of which don't pass tests and have somewhat unclear commit messages.  It looks like some of them are from 2014.  I realize that that's hardly your fault, but it does make trying to look at this as a whole, once it's merged a little painful.
Also, I know you put a lot of work into this, and it went back and forth until you settled on what you wanted.  I think after you did settle on it, it would have been cleaner to squash most of the original commits (squelching the back-and-forth) and gotten this down to ~1-3 commits.

The rails docs have a relatively short explanation of what I'm trying to say:  http://edgeguides.rubyonrails.org/contributing_to_ruby_on_rails.html#commit-your-changes

PR #210, opened 2018-02-26
Massively cleaned up Rakefile

luke-hill:
Ronseal PR. I also realised when I did the gemspec refactors I prevented people from Using Rake 12 and above, so I've just uncapped it (But left the minimum version at 11 for now).

I've removed the yard raketasks for now as they were never called. To be frank I contemplated removing yard entirely as it doesn't seem to have much use when our docs are already amazing and we enforce such strict guidelines on the repo already. Probably will remove this in a future PR and in a future release

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,15 +1,20 @@
 # frozen_string_literal: true
 
 task :cukes do
-  system 'bundle exec cucumber'
+  @cuke_exit_code = system 'bundle exec cucumber' || raise('Cukes failed')
 end
 
 task :specs do
-  system 'bundle exec rspec'
+  @spec_exit_code = system 'bundle exec rspec' || raise('Specs failed')
 end
 
 task :rubocop do
-  system 'bundle exec rubocop'
+  @cops_exit_code = system 'bundle exec rubocop' || raise('Cops failed')
 end
 
-task default: %i[rubocop specs cukes]
+task :default do
+  Rake::Task[:rubocop].invoke
+  Rake::Task[:specs].invoke
+  Rake::Task[:cukes].invoke
+  (@cuke_exit_code && @spec_exit_code && @cops_exit_code) || raise('Failed to pass all 3')
+end
--- a/Rakefile
+++ b/Rakefile
@@ -1,20 +1,15 @@
 # frozen_string_literal: true
 
 task :cukes do
-  @cuke_exit_code = system 'bundle exec cucumber' || raise('Cukes failed')
+  system 'bundle exec cucumber' || raise('Cukes failed')
 end
 
 task :specs do
-  @spec_exit_code = system 'bundle exec rspec' || raise('Specs failed')
+  system 'bundle exec rspec' || raise('Specs failed')
 end
 
 task :rubocop do
-  @cops_exit_code = system 'bundle exec rubocop' || raise('Cops failed')
+  system 'bundle exec rubocop' || raise('Cops failed')
 end
 
-task :default do
-  Rake::Task[:rubocop].invoke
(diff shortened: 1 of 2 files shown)


luke-hill:
@tgaff @TheMetalCode - Nice simple PR to review

luke-hill:
That's fine. I haven't used travis before. The way in which we catch it in my company is through a shell script that looks for an exit code. The exit code from these runs won't be (As you've hinted) 0.

I'll dive into travis docs.

luke-hill:
So what I've done is put up 2 potential fixes. The first fix is to catch failures at a unit level (Cuke / Spec / Cop), the second one catches them at a PR level.

Which would you guys prefer? I'm happy with either. Catching them at a Unit level is less code. But would exit as soon as something fails. Catching at a PR level is a bit more code, but would allow all 3 items to run for each PR? Thoughts?

tgaff:
@luke-hill Maybe you can explain to me why you want to make this change.  To my eyes the Rakefile looks fine as is and afaik is working fine.  The way it's setup now is pretty standard as well, that's why Cucumber and RSpec include those tasks.

Rakefile
```diff
@@ -1,30 +1,20 @@
...
+task :rubocop do
+  @cops_exit_code = system 'bundle exec rubocop' || raise('Cops failed')
 end
 
-task default: %w[rubocop spec:all cuke:all]
+task :default do
+  Rake::Task[:rubocop].invoke
+  Rake::Task[:specs].invoke
+  Rake::Task[:cukes].invoke
+  (@cuke_exit_code && @spec_exit_code && @cops_exit_code) || raise('Failed to pass all 3')
```
TheMetalCode: Is this line necessary? It looks like each task is already raising upon failure.

Rakefile
```diff
@@ -1,30 +1,20 @@
...
-require 'rubocop/rake_task'
-
-require 'yard'
-
-namespace :cuke do
-  Cucumber::Rake::Task.new(:all) do |t|
-    t.cucumber_opts = '--format pretty'
-  end
+task :cukes do
+  @cuke_exit_code = system 'bundle exec cucumber' || raise('Cukes failed')
```
TheMetalCode: Not sure we need to store the exit code in a var, do we? The `|| raise ('Cukes failed')` fulfills the purpose of stopping the build upon failure, right?

TheMetalCode:
I mean on the one hand, `bundle exec rubocop/rspec/cucumber` gives us the exact same functionality as the rake tasks, and they're a lot less verbose. The only think `rake` is really doing for us that the `bundle exec` shell commands don't is giving us a simple `rake` command to run the entire build locally. So this PR gives us the same benefit with quite a bit less code. I'm not outright opposed to it, especially now that it properly fails the Travis build when tests fail. At past jobs in a JavaScript context I've seen developers get sick of defining Gulp tasks for things they can just as easily issue shell commands for, so most Gulp tasks got replaced by shell commands being aggregated in `package.json`. This is a similar concept, it seems to me.

That said, the Rakefile wasn't that big to begin with and it probably wasn't going to get much bigger so I'm not sure it was screaming out to be pared down.

luke-hill:
I'm basically trying to diet the suite down and remove all wrapper commands and just have a clean API.

In the jobs I've been in I've never had the usage of `Cucumber::Rake::Task do` which is just doing block invocation. Under the hood both of those just do raw shell commands.

It's not a case of "wanting" to make these changes. It's a case of just doing a quick bit of tidy-up. That's all.

As we've had a good discussion about this and discussed the merits. I'm going to do a like for like replacement (Which will mean the removal of the final check)

PR #211, opened 2018-02-28
Addressable / Rubocop Generic /lib Tidy

luke-hill:
So this PR was intended to clear one of the tasks from the ever-growing TODO list but after a couple of days and not getting very far I've cut my losses.

This PR is non-breaking and does "some" refactoring around this class, which has way too much old style ruby for my liking.

I've also included a reduction to the `LineLength` metric. I didn't go all the way down to 80 to keep the PR reasonably small. But this at least does "some" of the required work.

In the 2nd commit SHA: https://github.com/natritmeyer/site_prism/commit/72248f1280970417b6817a701e939a38059f9996 I've explained what each individual small refactor is and how it's ok to do so

### DIFF
--- a/lib/site_prism/addressable_url_matcher.rb
+++ b/lib/site_prism/addressable_url_matcher.rb
@@ -5,7 +5,7 @@ require 'base64'
 
 module SitePrism
   class AddressableUrlMatcher
-    COMPONENT_NAMES = %w[scheme user password host port path query fragment].map(&:to_sym).freeze
+    COMPONENT_NAMES = %i[scheme user password host port path query fragment].freeze
     COMPONENT_PREFIXES = {
       query: '?',
       fragment: '#'
@@ -24,8 +24,9 @@ module SitePrism
       result = {}
       COMPONENT_NAMES.each do |component|
         component_result = component_matches(component, uri)
-        result = component_result ? result.merge!(component_result) : nil
-        break unless result
+        return nil unless component_result
+
+        result.merge!(component_result)
       end
       result
     end
@@ -35,7 +36,7 @@ module SitePrism
     def matches?(url, expected_mappings = {})
       actual_mappings = mappings(url)
       if actual_mappings
-        expected_mappings.empty? ? true : all_expected_mappings_match?(expected_mappings, actual_mappings)
+        expected_mappings.empty? || all_expected_mappings_match?(expected_mappings, actual_mappings)
       else
         false
       end
@@ -57,7 +58,6 @@ module SitePrism
       end
     end
 
-    # Memoizes the extracted component templates
     def component_templates
       @component_templates ||= extract_component_templates
     end
@@ -66,14 +66,14 @@ module SitePrism
       COMPONENT_NAMES.each_with_object({}) do |component, component_templates|
         component_url = to_substituted_uri.public_send(component).to_s
 
-        next unless component_url && component_url != ''
+        next unless component_url && !component_url.empty?
 
         reverse_substitutions.each_pair do |substituted_value, template_value|
           component_url = component_url.sub(substituted_value, template_value)
         end
 
         component_templates[component] = Addressable::Template.new(component_url.to_s)
-      end.freeze
+      end
     end
 
     # Returns empty hash if the template omits the component, a set of substitutions if the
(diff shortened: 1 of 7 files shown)


luke-hill:
@tgaff @TheMetalCode Probably less of an important PR. But a bit tricky

lib/site_prism/addressable_url_matcher.rb
```diff
@@ -66,14 +66,14 @@ def extract_component_templates
       COMPONENT_NAMES.each_with_object({}) do |component, component_templates|
         component_url = to_substituted_uri.public_send(component).to_s
 
-        next unless component_url && component_url != ''
+        next unless component_url && !component_url.empty?
```
tgaff: Is there ever a chance this is NOT a string (given the `.to_s` above)?   If not this could devolve to the simple `next if component_url.empty?`

lib/site_prism/addressable_url_matcher.rb
```diff
@@ -35,7 +36,7 @@ def mappings(url)
     def matches?(url, expected_mappings = {})
       actual_mappings = mappings(url)
       if actual_mappings
-        expected_mappings.empty? ? true : all_expected_mappings_match?(expected_mappings, actual_mappings)
+        expected_mappings.empty? || all_expected_mappings_match?(expected_mappings, actual_mappings)
```
tgaff: If you want to be overly clever you could do something like: `actual_mappings && (expected_mappings.empty? || all_expected_mappings_match?(expected_mappings, actual_mappings)) || false`  (but that's likely not so easy to grok.)

site_prism.gemspec
```diff
@@ -12,7 +12,8 @@ Gem::Specification.new do |s|
   s.email       = 'nat@natontesting.com'
   s.homepage    = 'http://github.com/natritmeyer/site_prism'
   s.summary     = 'A Page Object Model DSL for Capybara'
-  s.description = 'SitePrism gives you a simple, clean and semantic DSL for describing your site using the Page Object Model pattern, for use with Capybara'
+  s.description = 'SitePrism gives you a simple, clean and semantic DSL for describing your site.
+SitePrism implements the Page Object Model pattern, through its usage with Capybara.'
```
tgaff: "through its usage with Capybara." this fragment doesn't parse well for me.  I'm not sure that's the correct preposition (I expected "usage of Capybara" or "use with Capybara").  
Are you trying to say that SitePrism is used along-side (together with) Capybara or that SitePrism uses Capybara?

How about "SitePrism implements the Page Object Model pattern on top of Capybara." ?  Or "SitePrism implements the Page Object Model pattern for usage with Capybara."  (I prefer my 1st suggestion.)

spec/section_spec.rb
```diff
@@ -53,9 +53,11 @@ class PageWithSection < SitePrism::Page
...
 
       it 'should raise an ArgumentError' do
-        class Page < SitePrism::Page
-        end
-        expect { section }.to raise_error ArgumentError, 'You should provide section class either as a block, or as the second argument'
+        class Page < SitePrism::Page; end
+
+        expect { section }
+          .to raise_error(ArgumentError)
+          .with_message('You should provide section class either as a block, or as the second argument')
```
tgaff: trivial: should error messages end in a period?

lib/site_prism/addressable_url_matcher.rb
luke-hill: As far as I could tell from debugging (I spent a while on this), the tests sometimes have strings and sometimes don't. I didn't check which circumstances have which.

I think this ideally will serve as a "first refactor" with someone else with perhaps a bit more ability fixing this up

lib/site_prism/addressable_url_matcher.rb
luke-hill: Not going to lie, this one I actually didn't debug. I just know this code is semantically similar with one less conditional

spec/section_spec.rb
luke-hill: They should. Not sure why this passes. Will re-triage

luke-hill:
@tgaff You happy for this to hit master as is?

tgaff:
Yes go for it!  All my comments were optional (and one was purely
amusement).

PR #213, opened 2018-02-28
Clarify README for expected_elements

TheMetalCode:
A quick follow-up to #208.

NOTE: In reading about load validations via `Loadable`, I'm now somewhat concerned that `all_there?` modified by `expected_elements` is a tad obselete. When I originally submitted #98 (which is what inspired #208) I'm fairly sure `Loadable` wasn't implemented yet (or maybe it was and I didn't notice).

In any event, as long as we have the `expected_elements` construct, we'll want the README to be clear on how to use it.

### DIFF
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -24,7 +24,7 @@ module SitePrism
     alias collection elements
 
     def expected_elements(*elements)
-      @expected_items = elements.map(&:to_s)
+      @expected_items = elements
     end
 
     def section(section_name, *args, &block)
@@ -64,7 +64,7 @@ module SitePrism
 
     def add_to_mapped_items(item)
       @mapped_items ||= []
-      @mapped_items << item.to_s
+      @mapped_items << item
     end
 
     def raise_if_block(obj, name, has_block)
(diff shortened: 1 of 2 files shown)


luke-hill:
I'm happy with this, however I'm not sure how this solves what you were worried about (Loadable). As this just seems to add an extra element to the check.

If that's all you're wanting I'll approve and merge though. Let me know.

TheMetalCode:
@luke-hill Oh this is just a README change, so folks know how to use `expected_elements` when they want to specify more than one element, namely that it's comma-separated args and not an array. The musings about Loadable were a side-note: I'm concerned that `expected_elements` is not actually as useful as I thought. Those musings shouldn't necessarily block this PR.

lib/site_prism/element_container.rb
```diff
@@ -24,7 +24,7 @@ def elements(collection_name, *find_args)
     alias collection elements
 
     def expected_elements(*elements)
-      @expected_items = elements.map(&:to_s)
+      @expected_items = elements
```
TheMetalCode: @luke-hill This addresses your comment in #208. You're correct, the `to_s` is unnecessary.

PR #215, opened 2018-03-04
Support all selector types for iframe

### LINKED ISSUE #118: Scoping iframe by non-ID selectors
If I am reading http://www.rubydoc.info/github/jnicklas/capybara/Capybara%2FSelenium%2FDriver%3Awithin_frame correctly, it looks like Capybara is now able to scope an iframe by an actual element, instead of just an ID. Would it be possible for Site Prism to now allow mapping iframes by a more generic selector than just ID?

ricmatsui:
This PR fixes #118 by modifying the behavior of the iframe locator argument to use the CSS option for find when passed a string, and supporting all other options for locating an iframe:

```
iframe :my_iframe, MyIframe, '#the_iframe' # Use `:css` selector now instead of `:frame`
iframe :index_iframe, MyIframe, 0 # Use index selector as before
iframe :named_iframe, MyIframe, '[name=the_iframe]' # Use `:css` selector now
iframe :xpath_iframe, MyIframe, :xpath, '//iframe[@name="the_iframe"]' # Support `:xpath` selector
```

This also requires a update to the minimum version of Capybara to 2.12 where support for more selectors was added https://github.com/teamcapybara/capybara/commit/53b99af750277addae617268d4c57ad418fc63d1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1517,7 +1517,7 @@ end
 To expose the iframe, reference it from another page or class using the `iframe`
 method. The `iframe` method takes 3 arguments; the name by which you
 would like to reference the iframe, the page class that represents the
-iframe, and an ID or class by which you can locate the iframe. For example:
+iframe, and the CSS selector by which you can locate the iframe. For example:
 
 ```ruby
 class PageContainingIframe < SitePrism::Page
@@ -1525,8 +1525,21 @@ class PageContainingIframe < SitePrism::Page
 end
 ```
 
-The third argument to the `iframe` method must
-contain a selector that will locate the iframe node.
+### Locating an iframe
+
+While the above example uses a CSS selector to find the iframe, it is also
+possible to use an XPath expression or the index of the iframe in its parent
+(a shortcut for an `nth-of-type` CSS selector). For example:
+
+```ruby
+class PageContainingIframe < SitePrism::Page
+  # XPath Expression:
+  iframe :my_iframe, MyIframe, :xpath, '//iframe[@id="my_iframe_id"]'
+
+  # Index (nth-of-type) Selector:
+  iframe :my_iframe, MyIframe, 0
+end
+```
 
 ### Testing for an iframe's existence
 
--- a/features/iframes.feature
+++ b/features/iframes.feature
@@ -7,6 +7,17 @@ Feature: IFrame interaction
   Scenario: locate an iframe by index
     When I navigate to the home page
     Then I can locate the iframe by index
+    And I can see elements in an indexed iframe
+
+  Scenario: locate an iframe by name
+    When I navigate to the home page
+    Then I can locate the iframe by name
+    And I can see elements in a named iframe
+
+  Scenario: locate an iframe by xpath
+    When I navigate to the home page
+    Then I can locate the iframe by xpath
+    And I can see elements in an xpath iframe
 
   Scenario: interact with elements in an iframe
     When I navigate to the home page
--- a/features/step_definitions/iframe_steps.rb
+++ b/features/step_definitions/iframe_steps.rb
@@ -12,8 +12,38 @@ Then(/^I can locate the iframe by index$/) do
   expect(@test_site.home).to have_index_iframe
 end
 
+Then(/^I can locate the iframe by name$/) do
+  @test_site.home.wait_for_named_iframe
+
+  expect(@test_site.home).to have_named_iframe
+end
+
+Then(/^I can locate the iframe by xpath$/) do
+  @test_site.home.wait_for_xpath_iframe
+
+  expect(@test_site.home).to have_xpath_iframe
+end
+
 Then(/^I can see elements in an iframe$/) do
   @test_site.home.my_iframe do |f|
+    expect(f.some_text.text).to eq('Some text in an iframe')
+  end
+end
+
+Then(/^I can see elements in an indexed iframe$/) do
+  @test_site.home.index_iframe do |f|
+    expect(f.some_text.text).to eq('Some text in an iframe')
+  end
+end
+
+Then(/^I can see elements in a named iframe$/) do
+  @test_site.home.named_iframe do |f|
+    expect(f.some_text.text).to eq('Some text in an iframe')
+  end
+end
+
+Then(/^I can see elements in an xpath iframe$/) do
+  @test_site.home.xpath_iframe do |f|
     expect(f.some_text.text).to eq('Some text in an iframe')
   end
 end
--- a/test_site/pages/home.rb
+++ b/test_site/pages/home.rb
@@ -33,4 +33,6 @@ class TestHomePage < SitePrism::Page
   # iframes
   iframe :my_iframe, MyIframe, '#the_iframe'
   iframe :index_iframe, MyIframe, 0
+  iframe :named_iframe, MyIframe, '[name="the_iframe"]'
+  iframe :xpath_iframe, MyIframe, :xpath, '//iframe[@name="the_iframe"]'
 end
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -48,15 +48,15 @@ module SitePrism
       end
     end
 
-    def iframe(iframe_name, iframe_page_class, selector)
-      element_selector = deduce_iframe_element_selector(selector)
-      scope_selector = deduce_iframe_scope_selector(selector)
+    def iframe(iframe_name, iframe_page_class, *args)
+      element_find_args = deduce_iframe_element_find_args(args)
+      scope_find_args = deduce_iframe_scope_find_args(args)
       add_to_mapped_items iframe_name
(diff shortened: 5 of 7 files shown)


luke-hill:
Initially this seems good. However I'm unlikely to get to this for a couple of days. Thankyou for the contribution though. I'll get round to reviewing probably Thursday

features/step_definitions/iframe_steps.rb
```diff
@@ -12,12 +12,36 @@
...
+  expect(@test_site.home).to have_xpath_iframe
+end
+
 Then(/^I can see elements in an iframe$/) do
   @test_site.home.my_iframe do |f|
     expect(f.some_text.text).to eq('Some text in an iframe')
   end
 end
 
+Then(/^I can see elements in a named iframe$/) do
```
luke-hill: Can you refactor this step and all other similar steps (Including the ones that aren't yours), to remove the block. As it's superflous.

Here you would do: `@test_site.home.named_iframe.some_text.text`

features/iframes.feature
```diff
@@ -8,6 +8,16 @@ Feature: IFrame interaction
     When I navigate to the home page
     Then I can locate the iframe by index
```
luke-hill: Can we add the extra cuke here (Will need writing).

Don't need to do it for the id one as the final test duplicates that - I need to check all this stuff over later. But this definitely can get added now.

test_site/pages/home.rb
```diff
@@ -33,4 +33,6 @@ class TestHomePage < SitePrism::Page
   # iframes
   iframe :my_iframe, MyIframe, '#the_iframe'
   iframe :index_iframe, MyIframe, 0
+  iframe :named_iframe, MyIframe, '[name=the_iframe]'
```
luke-hill: `"the_frame"`

features/step_definitions/iframe_steps.rb
ricmatsui: @luke-hill I tried without the block, but it does not work because as far as I know the block is necessary since Capybara needs to switch to the frame before being able to examine the elements within the frame:
```
  Scenario: locate an iframe by name         # features/iframes.feature:11
    When I navigate to the home page         # features/step_definitions/navigation_steps.rb:3
    Then I can locate the iframe by name     # features/step_definitions/iframe_steps.rb:15
    And I can see elements in a named iframe # features/step_definitions/iframe_steps.rb:33
      undefined method `call' for nil:NilClass
      Did you mean?  caller (NoMethodError)
      ./lib/site_prism/element_container.rb:60:in `block (2 levels) in iframe'
      ./lib/site_prism/element_container.rb:59:in `block in iframe'
      ./features/step_definitions/iframe_steps.rb:34:in `/^I can see elements in a named iframe$/'
      features/iframes.feature:14:in `And I can see elements in a named iframe'
```

As mentioned in the docs:

> Due to capybara internals it is necessary to pass a block to the iframe instead of simply calling methods on it; the block argument is the SitePrism::Page that represents the iframe's contents.

https://github.com/natritmeyer/site_prism/blame/master/README.md#L1557-L1561

Is there some other way I should try?

features/iframes.feature
ricmatsui: @luke-hill Can you explain more about the extra cuke I should add? Do you mean I should be adding these steps?
```
Then I can see elements in an iframe with capybara query options
And I can see an expected bit of text
```

features/step_definitions/iframe_steps.rb
luke-hill: Nope that's fine. I've not used iFrames much in my day-to-day. And the way I use it is with capybaras `within_iframe do` which uses a block. So it makes sense for the evaluation in SP to follow suit. Sorry for that.

luke-hill:
Remaining Update points

1) Code wise I'm happy: Some stylistic stuff perhaps needs tweaking but I see you've kept to the style of the codebase so that's fine (I think it needs updating).

2) I have a slight concern about unit test coverage (The different use cases), albeit these are in private methods. I've done a quick scan of the specs and can't find any specs for iFrames. So I won't enforce you to change these. As you've got Cucumber Tests I'm less inclined to push for this yet (Until we've done a full code-sweep)

3) What I will ask you do before this is ready to release is add some notes to the README about how to create the different iFrames. We need some documentation around this functionality ideally. If there is existing stuff in the Readme feel free to update as you see fit.

ricmatsuitile:
@luke-hill Sounds good, I have add the cuke and included additional notes in the README now, thanks: https://github.com/natritmeyer/site_prism/pull/215/commits/16891199ce3ad55ca855c2bc975158b6254635e2#diff-04c6e90faac2675aa89e2176d2eec7d8

tgaff:
This looks good, and thank you!

The test for `index_iframe` could be a little more robust if there was more than one iframe and/or a counter example test trying to access an index out of bounds.

Also...
It's obviously allowing passing parameters through to scope the iframe further (e.g. `@test_site.home.has_iframe2?(class: 'zork')`).  I don't see a test for that for the iframe specifically, although there are tests for this feature on regular `element`s.  Would it make sense to add one?  Also similar to that, I note that we don't support scoping like that when calling into the iframe itself.  

```rb

# this currently is supported
@test_site.home.my_iframe do |f|
  expect(f.some_text.text).to eq('Some text in an iframe')
end

# this is too
@test_site.home.has_my_iframe?(id: 'facebook') #note the id scoping

# but this is not
@test_site.home.my_iframe(id: 'facebook') do |f|
  expect(f.some_text.text).to eq('Some text in an iframe')
end
```

Probably not necessary for this PR, but potential future development.  @luke-hill ?

README.md
```diff
@@ -1528,6 +1528,25 @@ end
 The third argument to the `iframe` method must
```
luke-hill: lets remove these 2 lines. We don't use them elsewhere

README.md
```diff
@@ -1528,6 +1528,25 @@ end
...
 
+### Locating an iframe
+
+While the above example uses a CSS selector to find the iframe, it is also
+possible to use an XPath expression or the index of the iframe in its parent
+(a shortcut for an `nth-of-type` CSS selector). For example:
+
+```ruby
+class PageContainingIframe < SitePrism::Page
+  # CSS Selector:
```
luke-hill: remove these three lines

luke-hill:
I'm happy to merge this with a couple of additions to the TODO docs.

- Review iFrame specs, add/amend where appropriate
- Allow scoping iframes to then be passed into element native object

ricmatsui:
@luke-hill Sounds good, I have cleaned up the README now

luke-hill:
Hi @ricmatsui Don't worry about the conflicts yet. Wait until I re-ping you (Saves doing it multiple times).

It'll probably be sometime near the end of March / Early April.

luke-hill:
Hi @ricmatsui As a couple of other PR's have fell by the wayside and I'm focusing on dev based updates at the moment this is now more relevant to the current HEAD (And 2.12 cut),

As such I've removed the Next Release Candidate - As it can hit this release candidate. Can you fix up the issue (Just will be a capybara version conflict), and then I'll merge it in ASAP

Luke

ricmatsui:
@luke-hill It seems like the dependencies of Capybara 2.12 actually require Ruby >= 2.2.2 which is causing issues with Ruby 2.0/2.1. I think this means `site_prism` will have to increase its minimum Ruby version to >= 2.2.2, but I'm not sure. How should we handle this?

https://travis-ci.org/natritmeyer/site_prism/builds/352562755?utm_source=github_status&utm_medium=notification

tgaff:
Hopefully Luke Hill won't mind if I jump in here again.  Our timezones seem to be different enough that he's likely offline by now.  

I'm not sure that it's really the ruby version causing this error.  I was able to successfully run 3496d454ad6b82cf5693676da66e9f00e18af26b locally with ruby 2.1.3.  
And on travis (fork) here: https://travis-ci.org/tgaff/site_prism/builds/352590872

Capybara 2.12.1 sets a minimum ruby version of 1.9.3.  It looked like travis here was getting stuck because it pulled in xpath 3.0, but capybara 2.12.1 requires `s.add_runtime_dependency("xpath", ["~> 2.0"])`

You might try just rebasing on master and see if that helps.

ricmatsuitile:
@tgaff Thanks, that's strange, I'll try again. This branch is already rebased on master.

ricmatsui:
@tgaff Thanks that worked, not sure why it pulled incompatible dependencies the first time Travis built it maybe there's a better way to control with with `Gemfile.lock` for each Ruby version: https://travis-ci.org/natritmeyer/site_prism/builds/352562755

luke-hill:
To add to this, when I did a PR earlier - I got an invalid version of xpath on the very first build (2.0.0 firefox), but then everything else worked.

I'm going to couple my phantom failure with yours, and put it down to travis' black box?

PR #216, opened 2018-03-06
Recommended minimum ruby update for gem (2.0 -> 2.1.7)

luke-hill:
Self explanatory PR. Not sure "when" this should land. Definitely not next RC, possibly one after

@tgaff Should hopefully alleviate some of your concerns.
@TheMetalCode can you thumb if happy


tgaff:
I'd be in no hurry to force users of the gem to move up (aside from the fact that security patches ended over a year ago).  Continuing to allow the use of 2.0 has afaik no cost at the moment.  (Unless there's a 2.1 feature you're looking to take advantage of?)

My concern on the other PR was actually with **forcing potential developers** of the gem to use 2.0 for development work; not with **allowing users** of the gem to continue using 2.0 if their project is for some reason constrained to do so.

If you are going to stop supporting it though, I think it makes sense to delete the travis tests for 2.0.0.  (looks like the travis failure agrees)
👍on the 2.4.0 tests, maybe add 2.5.0 also!

luke-hill:
Was more a case of just having the PR ready. Essentially it's complaining about running ruby 2.0 through our tests when we specify a minimum version of 2.1 - I'll have a quick re-think about how to do it.

Do we know as a community how many people are using Ruby 2.5 yet? I know I've not even touched it personally. Are there any usage stats anywhere?

luke-hill:
@tgaff I believe this is now fixed. Are you happy for this to be approved and hit the next release (Not merging anything more into current release as it's fairly chunky with @TheMetalCode's big change with the `all_there?` refactor.

tgaff:
Looks great.  GTM

Thank you!

PR #214, opened 2018-03-04
Add chrome in Travis ("Next Up" No.2 of TODO.md)

RustyNail:
Hi, @luke-hill.
I have implemented the following tasks of TODO list.

```
2) Travis:
  - Add in Chrome
  - Update Driver/Browser
```

## Change point
- .travis.yml
  - Added chrome and its driver.
  - Update geckodriver to latest version.
- features/support/env.rb
  - Make profile according to browser.
- TODO.md
  - Deleted items implemented by this Pullrequest.

### Ref
- Setting up Google Chrome with travis
  - https://docs.travis-ci.com/user/chrome
- Latest geckodriver
  - https://github.com/mozilla/geckodriver/releases/

PR #217, opened 2018-03-07
Travis chrome update

luke-hill:
Repurposed version of https://github.com/natritmeyer/site_prism/pull/214

Added 2nd commit where we don't break local development (Have added a side note that we need to update "how" to test and develop locally)

TL;DR for the :point_up: is you need to run this locally now for any future development
```
$ cp .env.example .env
```


features/support/env.rb
```diff
@@ -43,13 +45,31 @@
...
+  else
+    Selenium::WebDriver::Firefox::Profile.new.tap do |profile|
+      profile['browser.cache.disk.enable'] = false
+      profile['browser.cache.memory.enable'] = false
+    end
+  end
+end
+
+def browser
+  ENV['browser'].to_sym
```
tgaff: Maybe you could just set a default here if the environment variable is unset?  That would make it easier for most users.

luke-hill: Agreed. I think in one of the commits I didn't copy over he specified it to default to firefox. Will add that in (Along with a rebase)

Check back say Monday?

PR #220, opened 2018-03-09
TODO update / Generic House-Clean

luke-hill:
- Cleared a few tasks off the list
- Added / Removed TODO items
- Cleanup: Removed all traces of yard/redcarpet
- Added version spec primitive
- page and section specs amended - (Cleaned up some code, removed some duplicate specs, shaved a couple of seconds off runtime)

P.S. Ignore the commit name / message, it's just compiled of several commits rebased into one.

@tgaff @TheMetalCode to review (Kinda, there's nothing much here really)

### DIFF
--- a/spec/version_spec.rb
+++ b/spec/version_spec.rb
@@ -5,5 +5,5 @@ require 'spec_helper'
 describe 'SitePrism::VERSION' do
   subject { SitePrism::VERSION }
 
-  it { is_expected.to eq('2.11') }
+  it { is_expected.to be_truthy }
 end
--- a/spec/version_spec.rb
+++ b/spec/version_spec.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'SitePrism::VERSION' do
+  subject { SitePrism::VERSION }
+
+  it { is_expected.to eq('2.11') }
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,6 @@
-doc/*
-.yardoc/*
 .DS_Store
 Gemfile.lock
 coverage
 .bundle
-.ruby-gemset
 .idea
-.todo-list
 .env
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,1 +1,0 @@
-site_prism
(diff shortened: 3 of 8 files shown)


spec/version_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'SitePrism::VERSION' do
+  subject { SitePrism::VERSION }
+
+  it { is_expected.to eq('2.11') }
```
TheMetalCode: I'm somewhat skeptical of the value of this test. It basically says "Ruby has class constants and they work" which isn't a thing site_prism is really responsible for, plus this is a test that you'll definitely have to update every time you cut a release. Tests can sometimes be valuable just for the sake of documentation, but even on that score if I wanted to know the version I'd check the gemspec or, absent that, whatever file was defining the `VERSION` constant. It's not something I'd go looking for in the rspec tests.

I wouldn't block the PR over this - it's not a _lot_ of extra maintenance added when bumping version. Just not sure it's really necessary is all.

TheMetalCode: I'm somewhat skeptical of the value of this test. It basically says "Ruby has class constants and they work" which isn't a thing site_prism is really responsible for, plus this is a test that you'll definitely have to update every time you cut a release. Tests can sometimes be valuable just for the sake of documentation, but even on that score if I wanted to know the version I'd check the gemspec or, absent that, whatever file was defining the VERSION constant. It's not something I'd go looking for in the rspec tests.

I wouldn't block the PR over this - it's not a lot of extra maintenance added when bumping version. Just not sure it's really necessary is all.

tgaff: Perhaps just verify the constant is set.

luke-hill: I think actually I'd agree with what you've said, but it's there for precisely that reason. Just to add extra compliance to a codebase I'm not yet fully familiar with. It also adds a spec which would stop accidental versions being bumped.

But having said that, you've both come onboard to review it, which I'm greatful for, as it gives more than just my eyes on it. So I'll alter it to `be_truthy` sometime in the next couple of days.

PR #221, opened 2018-03-12
Dev Update/Bump

luke-hill:
This was originally going to land as 1 PR but I've split it out into 2 and branched from this one for the other one in my dev environment.

The purpose was to provide a cleanup of a load of redundant options inside the cucumber tests - Saving a small amount of time but mainly just improving readability (Also because a lot are redundant), as well as a slight bump to our gem dependencies - Specifically Selenium Webdriver.

This PR focuses just on the dependencies, the next one will tackle some of the cukes - As one config option was masking a lot of poor setup!

All of the items removed are either default / redundant. The Profiles have been removed because Chrome's did nothing (It was just there for backwards compatibility), and Firefox's only did some basic cache removal. Which creating new Boxes each time would render redundant - The only possibly caveat to this is that it might affect local dev runs, however now we're using an up to date version of geckodriver (And local developers should too!), this problem shouldn't occur


luke-hill:
ping @tgaff @TheMetalCode

PR #222, opened 2018-03-12
Cucumber / Step Cleaner

luke-hill:
Removed quite a few duplicate / superfluous specs that were basically doing the same thing several times over.

Tried to establish a standard of not wanting to test a parent item for existence and then a child item for anything (As we must have an existent parent in order to interrogate the child) <- This also will essentially fail with `NoMethodError` but on a very specific call otherwise, which is the same as what we had before, just not in RSpec territory.

Also started to DRY up some suite loading stuff using Cucumber Hooks. For now I've just put in a basic creation, I've yet to memoize anything, I wasn't wanting to do anything too drastic just yet.

I've also just generally tried to make the overall feel much clearer with regards to naming e.t.c. There still is more work to do especially around breaking the code out into more manageable chunks but for now this does a good interim job of making the suite run a little faster, a bit more concise, and with fewer overlap points.

### DIFF
--- a/features/page_element_interaction.feature
+++ b/features/page_element_interaction.feature
@@ -32,17 +32,17 @@ Feature: Page element interaction
 
   Scenario: Wait for element
     When I navigate to the home page
-    Then when I wait for the element that takes a while to appear
-    Then I successfully wait for it to appear
+    And I wait for the element that takes a while to appear
+    Then the slow element appears
 
   Scenario: Wait specific amount of time for element to appear
     When I navigate to the home page
-    And I wait for a specifically short amount of time for an element to appear
+    And I wait for a short amount of time for an element to appear
     Then the element I am waiting for doesn't appear in time
 
-  Scenario: Check that all elements are present
+  Scenario: Page without `expected_elements`
     When I navigate to the home page
-    Then all expected elements are present
+    Then not all expected elements are present
 
   Scenario: Page with `expected_elements`
     When I navigate to the home page that contains expected elements
--- a/features/page_element_interaction.feature
+++ b/features/page_element_interaction.feature
@@ -69,7 +69,7 @@ Feature: Page element interaction
 
   Scenario: Wait for invisibility of element
     When I navigate to the home page
-    And I wait while for an element to become invisible
+    And I wait for an element to become invisible
     Then the previously visible element is invisible
 
   Scenario: Wait specific amount of time for invisibility of element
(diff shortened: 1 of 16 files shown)


luke-hill:
ping @tgaff @TheMetalCode

features/page_element_interaction.feature
```diff
@@ -32,17 +32,17 @@ Feature: Page element interaction
 
   Scenario: Wait for element
     When I navigate to the home page
```
TheMetalCode: So this feature is interesting because 16 of 18 scenarios use `When I navigate to the home page`, so at least visually you'd see some benefit from a Background step here.

But then there are those 2 scenarios that would have to load a page after the home page is loaded. In this particular suite we're loading small, simple html pages so I don't think there'd be a performance hit. In the real world though, you really wouldn't want to double-load. And there are folks that are going to use this suite as an example of what to do in their own work.

Ultimately I'd probably add the Background step here for ease of maintenance.

luke-hill: Both this one and the other (plural) feature need splitting out better. It feels (And the specs do as well), like a proper melting pot. But this PR already had a sizeable amount of refactors and I didn't want to overload the PR.

Are you happy for me to tackle this and the other one (plural), in a separate PR, where we look to split them out into 2/3 different features each?

TheMetalCode: That seems reasonable, yes.

features/step_definitions/page_section_steps.rb
```diff
@@ -44,52 +45,44 @@
 
 Then(/^that section is there too$/) do
```
tgaff: This one is vague / strangely named if you want to add that to your refactor.  It's used in:
```gherkin

  Scenario: Designate a section of a page
    Then I can see elements in the section
    When I navigate to another page
    Then that section is there too

```

features/step_definitions/navigation_steps.rb
```diff
@@ -1,25 +1,33 @@
...
 When(/^I navigate to the redirect page$/) do
-  @test_site = TestSite.new
   @test_site.redirect_page.load
 end
 
+When(/^I navigate to a page with no title$/) do
+  @test_site.no_title.load
+end
+
+When(/^I navigate to another page$/) do
```
tgaff: Vague and related to my other comment.  This isn't being used as just **any** other page, it's being used as specifically the page_with_people page, so it might make sense to just say that.

PR #223, opened 2018-03-15
New Feature: Section#native

luke-hill:
This PR cleans up the `section_spec.rb` into the correc format as well as exposing a new public method which essentially just wraps `#native` when inside a section. This was a request from a while back just to be rehonoured.

I anticipate some refactoring required around the object_double which I've not used before and I'm sure there's a more elegant way to spoof a `Capybara::Node::Element` than the way I've done it.

ping @tgaff @TheMetalCode

### DIFF
--- a/spec/section_spec.rb
+++ b/spec/section_spec.rb
@@ -2,45 +2,41 @@
 
 require 'spec_helper'
 
+class Section < SitePrism::Section; end
+class Page < SitePrism::Page; end
+
 describe SitePrism::Page do
   describe '.section' do
-    class Page < SitePrism::Page; end
-    class Section < SitePrism::Section; end
-
-    class PageWithAnonymousSection < SitePrism::Page
-      section :anonymous_section, '.section' do |s|
-        s.element :title, 'h1'
-      end
-    end
-
-    class PageWithSection < SitePrism::Page
-      section :section, Section, '.section'
-    end
-
     context 'second argument is a Class' do
-      subject { PageWithSection.new }
-
-      it 'should create a section using the Class' do
-        expect(subject).to respond_to(:section)
+      class PageWithSection < SitePrism::Page
+        section :section, Section, '.section'
       end
+
+      subject(:page_with_section) { PageWithSection.new }
+
+      it { is_expected.to respond_to(:section) }
     end
 
     context 'second argument is not a Class and a block given' do
-      subject { PageWithAnonymousSection.new }
-
-      it 'should create an anonymous section using the block' do
-        expect(subject).to respond_to(:anonymous_section)
+      class PageWithAnonymousSection < SitePrism::Page
+        section :anonymous_section, '.section' do |s|
+          s.element :title, 'h1'
+        end
       end
+
+      subject(:page_with_anonymous_section) { PageWithAnonymousSection.new }
+
+      it { is_expected.to respond_to(:anonymous_section) }
     end
 
     context 'second argument is not a Class and no block given' do
-      subject { Page.section(:incorrect_section, '.section') }
+      subject(:invalid_page) { Page.section(:incorrect_section, '.section') }
       let(:error_message) do
         'You should provide section class either as a block, or as the second argument.'
       end
 
-      it 'should raise an ArgumentError' do
-        expect { subject }
+      it 'raises an ArgumentError' do
+        expect { invalid_page }
           .to raise_error(ArgumentError)
           .with_message(error_message)
       end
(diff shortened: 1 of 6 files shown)


spec/section_spec.rb
```diff
@@ -4,37 +4,56 @@
...
-              s.element :title, 'h1'
-            end
-          end
-
-          page = PageWithSection.new
-          expect(page).to respond_to(:anonymous_section)
-        end
+      subject { PageWithAnonymousSection.new }
+
+      it 'should create an anonymous section using the block' do
```
tgaff: I noticed that the SitePrism::Page tests at the top and the SitePrism::Section specs at the bottom use different naming styles.  Here tests are named "should create an anonym..." whereas below verbs are used without should (ie: "creates an anonym...")   
There's a big debate about which is better, but since you're editing most of this file anyhow you might consider making them consistent.

spec/section_spec.rb
```diff
@@ -4,37 +4,56 @@
 
 describe SitePrism::Page do
   describe '.section' do
+    class Page < SitePrism::Page; end
+    class Section < SitePrism::Section; end
+
+    class PageWithAnonymousSection < SitePrism::Page
```
tgaff: Each of these is only used in one of the contexts below.  It might make sense to move them down to the context in which they're relevant.

spec/section_spec.rb
tgaff: The ones above could maybe read better to especially around `context`.  Right now you'd get output like: "SitePrism::Page .section second argument is a Class should respond to #section" which while parseable is not the clearest.

spec/section_spec.rb
```diff
@@ -44,24 +63,71 @@ class Page < SitePrism::Page; end
...
 
-  it 'passes a given block to Capybara.within' do
-    expect(Capybara).to receive(:within).with('div')
+  describe 'instance' do
+    it 'passes a given block to Capybara.within' do
+      expect(Capybara).to receive(:within).with(locator)
 
-    SitePrism::Section.new(a_page, 'div') { 1 + 1 }
-  end
+      section_with_block
```
tgaff: This seems like it's the subject in this test.  Maybe consider splitting this into contexts with/without blocks? Or just move that code from the let down here.

spec/section_spec.rb
```diff
@@ -44,24 +63,71 @@ class Page < SitePrism::Page; end
...
+    end
 
-  it 'does not require a block' do
-    expect(Capybara).not_to receive(:within)
+    it 'does not require a block' do
+      expect(Capybara).not_to receive(:within)
 
-    SitePrism::Section.new(a_page, 'div')
-  end
+      subject
```
tgaff: Named subjects would be nicer.  

I also prefer to assert afterwards.  You could add a spy on `within` in a before block and then move your asserts to the end for several of these tests.  Personal preference though.

spec/section_spec.rb
```diff
@@ -44,24 +63,71 @@ class Page < SitePrism::Page; end
...
+    it 'evaluates visibility by delegating through root_element' do
+      expect(locator).to receive(:visible?)
 
-    it 'responds to javascript methods' do
-      expect(section).to respond_to(:execute_script)
-      expect(section).to respond_to(:evaluate_script)
+      subject.visible?
+    end
+
+    it 'obtains the text of a section by delegating through root_element' do
```
tgaff: Thank you for adding these tests!

Since you're testing separate public methods of Section here (the API) it might be clearer to name your tests to make that super obvious.  
e.g.

```
describe '#visible?' do
  it 'delegates through root_element'
```

If you feel that's too long, I won't object to combining the two lines into one for now.  I do think this makes it clearer that *this test* is the test for the `#visible` method.

spec/sections_spec.rb
```diff
@@ -19,11 +19,11 @@ class APage < SitePrism::Page
     end
 
     it 'should create matching object method' do
-      expect(page).to respond_to(:single_section)
+      expect(subject).to respond_to(:single_section)
```
tgaff: Nagging you about this one last time....If you read my notes above you'll already know that I am strongly in favor of using named subjects.  (The RSpec relish docs seem to point in that direction as well.)  Please consider changing this.

spec/sections_spec.rb
```diff
@@ -19,11 +19,11 @@ class APage < SitePrism::Page
...
     end
 
     it 'should create matching existence method' do
-      expect(page).to respond_to(:has_single_section?)
+      expect(subject).to respond_to(:has_single_section?)
```
tgaff: Maybe I'll look into a way to better implement/test these.  This only verifies the method is created, I think we rely entirely on cukes for validating the method behavior. 😞

spec/section_spec.rb
```diff
@@ -4,37 +4,56 @@
 
 describe SitePrism::Page do
```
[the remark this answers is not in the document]
TheMetalCode: I'm not sure what you're referring to here. Do you mean the describe block should be something other than `SitePrism::Page`?

spec/section_spec.rb
```diff
@@ -4,37 +4,56 @@
...
+
+    class PageWithSection < SitePrism::Page
+      section :section, Section, '.section'
+    end
+
+    context 'second argument is a Class' do
+      subject { PageWithSection.new }
+
+      it 'should create a section using the Class' do
+        expect(subject).to respond_to(:section)
```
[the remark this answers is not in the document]
TheMetalCode: I think unnamed subjects are fine if they're being declared close to the assertion. But if there's a lot of distance between the subject declaration and the assertion, it's a readability problem because it's no longer easy to know what `subject` means.

But also, in this case you don't actually need an explicit subject declaration so I'd agree with just using a simple `it` block as @tgaff suggests.

spec/section_spec.rb
TheMetalCode: +1 for consistency, and for making sure the spec output reads plainly.

I come down against `should` and in favor of making strong declarations with it blocks. http://www.betterspecs.org/#should

spec/section_spec.rb
```diff
@@ -4,37 +4,56 @@
...
-          .with_message('You should provide section class either as a block, or as the second argument.')
+          .with_message(error_message)
       end
     end
   end
 end
 
 describe SitePrism::Section do
-  let(:a_page) { class Page < SitePrism::Page; end }
+  class Page < SitePrism::Page; end
```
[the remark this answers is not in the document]
TheMetalCode: I mean the parent class of a Section being a SitePrism::Page is definitely the assumed usage even if it may not be strictly necessary for this group of tests.

spec/section_spec.rb
TheMetalCode: Yes, it would be a good idea to, in the near term, refactor specs so that the describe blocks _explicitly_ state which public API method they are testing. http://www.betterspecs.org/#describe

spec/sections_spec.rb
TheMetalCode: Furthermore, each of this it blocks can be easily made more succinct:

```ruby
it { is_exected.to respond_to(:single_section) }
it { is_expected.to respond_to(:has_single_section?) }
it { is_expected.to respond_to(:plural_sections) }
it { is_expected.to respond_to(:has_plural_sections?) }
```

TheMetalCode: Now if you're using these succinct it blocks, that probably reduces the utility of a named subject. But then I'd argue that your spec would be short enough that an unnamed subject would no longer be problematic (i.e. it wouldn't be hard to know what the subject is for each assertion).

spec/section_spec.rb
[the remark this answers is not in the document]
tgaff: We're in the SitePrism::**Section** spec, but describing SitePrism::**Page**.  I understand why it was done, but it feels a little awkward for a unit test.  ¯\_(ツ)_/¯

PR #224, opened 2018-03-16
Dev Dep Bump - Take 2

luke-hill:
Primary reason is to remove `config.match = :prefer_exact` - However in doing so I had to fix some styling on some of the test page's. Nothing major.

Added in a couple more tests and moved some items around. All fairly self explanatory - This allows all of the integration tests to pass and fail correctly instead of masking invalid selector errors.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright (c) 2011-2016, Nathaniel Ritmeyer
+Copyright (c) 2011-2018, Nathaniel Ritmeyer
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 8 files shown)


luke-hill:
Ping @tgaff @TheMetalCode

tgaff:
What does "dev dep bump" mean?  I had guessed development dependency bump, but this looks like it's all about cuke improvements.
Looks OK otherwise, obviously you'll want to investigate the test failure.

luke-hill:
Hi @tgaff - The dev dep bump is to update the development working processes (Which was here just tweaking one line).

In tweaking that 1 line a lot of stuff breaks - Because it was originally coded poorly #noblameculture - And expected these guards to cover it.

The cuke changes are nearly all necessitated by the tweak to the selector

Wil rebase and look into any failures.

Issue #80, opened 2014-07-10
Deprecate wait_for_ matcher


abotalov:
I propose to deprecate `wait_for_*` method.

Reasons:
- it violates standard practice to end methods that return boolean values with ?
- it does exactly the same what `has_*` method does

SitePrism exposes design that is very similar to Capybara's design (I mean has_, has_no, have). wait_for just doesn't fit. There's no use case when it may be preferable to has_*?

PR #226, opened 2018-03-17
New Feature: option for raising exception for wait timeout

ricmatsui:
This PR adds a new option to raise exceptions if a `wait_for_*` does not find the element waited for in the given timeout period or if a `wait_for_no_*` still finds the element it is waiting to not exist after the given timeout period. This will be off by default, but can be turned on in configuration. Raising exceptions as soon as a problem is detected is important to be able to debug when a test fails.

This was previously discussed in https://github.com/natritmeyer/site_prism/issues/80#issuecomment-48780149

Before:
```
raise WaitFailed unless wait_for_menu 10
raise WaitFailed unless wait_for_search_field 10
raise WaitFailed unless wait_for_search_results 10, count: 5

raise WaitFailed unless wait_for_no_menu 10
```

After:
```
SitePrism.raise_on_wait_fors = true

wait_for_menu 10
wait_for_search_field 10
wait_for_search_results 10, count: 5 

wait_for_no_menu 10
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1642,6 +1642,37 @@ with this:
 @search_page.search_results
 ```
 
+## Raising Errors on wait_for
+
+By default, when using `wait_for_*` methods, SitePrism will not raise
+an error if an element does not appear within the timeout period and will
+simply return `false` and allow the test to continue. This is different from
+the other methods such as `wait_until_*_visible` which do raise errors.
+
+Add the following code your spec_helper file to enable errors to be
+raised immediately when a `wait_for_*` does not find the element it is
+waiting for:
+
+```ruby
+SitePrism.configure do |config|
+  config.raise_on_wait_fors = true
+end
+```
+
+This enables you to replace this:
+
+```ruby
+raise unless @search_page.wait_for_search_results
+```
+
+with this:
+
+```
+# With raise on wait_fors enabled, this will automatically raise
+# if no search results are found
+@search_page.wait_for_search_results
+```
+
 ## Using SitePrism with VCR
 
 There's a SitePrism plugin called `site_prism.vcr` that lets you use
--- a/README.md
+++ b/README.md
@@ -1689,14 +1689,15 @@ with this:
 
 ## Raising Errors on wait_for
 
-By default, when using `wait_for_*` methods, SitePrism will not raise
-an error if an element does not appear within the timeout period and will
+By default, when using `wait_for_*` and `wait_for_no_*` methods, SitePrism will not raise
+an error if an element does not appear or disappear within the timeout period and will
 simply return `false` and allow the test to continue. This is different from
 the other methods such as `wait_until_*_visible` which do raise errors.
 
 Add the following code your spec_helper file to enable errors to be
-raised immediately when a `wait_for_*` does not find the element it is
-waiting for:
+raised immediately when a `wait_for_*` method does not find the element it is
+waiting for and when a `wait_for_no_*` method continues to find an element it is
+waiting to not exist:
 
 ```ruby
 SitePrism.configure do |config|
@@ -1708,6 +1709,8 @@ This enables you to replace this:
 
 ```ruby
 raise unless @search_page.wait_for_search_results
+# or...
+raise unless @search_page.wait_for_no_search_results
 ```
 
 with this:
@@ -1716,6 +1719,8 @@ with this:
 # With raise on wait_fors enabled, this will automatically raise
 # if no search results are found
 @search_page.wait_for_search_results
+# or automatically raise if search results are still found
+@search_page.wait_for_no_search_results
 ```
 
 ## Using SitePrism with VCR
--- a/features/step_definitions/wait_steps.rb
+++ b/features/step_definitions/wait_steps.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+Given(/^exceptions are configured to raise on wait_fors$/) do
+  SitePrism.raise_on_wait_fors = true
+end
+
 When(/^I wait for the element that takes a while to appear$/) do
   @test_site.home.wait_for_some_slow_element
 end
@@ -12,6 +16,12 @@ When(/^I wait for a short amount of time for an element to appear$/) do
   @test_site.home.wait_for_some_slow_element(1)
 end
 
+Then(/^an exception is raised when I wait for an element that won't appear$/) do
+  expect { @test_site.home.wait_for_some_slow_element(1) }
+    .to raise_error(SitePrism::TimeOutWaitingForExistenceError)
+    .with_message('Timed out after 1s waiting for TestHomePage#some_slow_element')
+end
+
 Then(/^the element I am waiting for doesn't appear in time$/) do
   expect(@test_site.home).not_to have_some_slow_element
 end
@@ -22,4 +32,10 @@ end
 
 Then(/^the slow section appears$/) do
   expect(@test_site.section_experiments.parent_section).to have_slow_section_element
+end
+
+Then(/^an exception is raised when I wait for a section that won't appear$/) do
+  expect { @test_site.section_experiments.parent_section.wait_for_slow_section_element(1) }
+    .to raise_error(SitePrism::TimeOutWaitingForExistenceError)
+    .with_message('Timed out after 1s waiting for Parent#slow_section_element')
 end
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -73,6 +73,13 @@ module SitePrism
       raise SitePrism::UnsupportedBlock, "#{obj.class}##{name}"
     end
 
+    def raise_wait_for_if_failed(obj, name, timeout, failed)
+      return unless SitePrism.raise_on_wait_fors && failed
+
+      raise SitePrism::TimeOutWaitingForExistenceError, \
+            "Timed out after #{timeout}s waiting for #{obj.class}##{name}"
+    end
+
     private
 
     def build(name, *find_args)
@@ -130,9 +137,11 @@ module SitePrism
       create_helper_method(method_name, *find_args) do
         define_method(method_name) do |timeout = nil, *runtime_args|
           timeout = timeout.nil? ? Capybara.default_max_wait_time : timeout
-          Capybara.using_wait_time(timeout) do
+          result = Capybara.using_wait_time(timeout) do
             element_exists?(*find_args, *runtime_args)
           end
+          self.class.raise_wait_for_if_failed(self, element_name.to_s, timeout, !result)
+          result
         end
       end
     end
--- a/lib/site_prism/exceptions.rb
+++ b/lib/site_prism/exceptions.rb
@@ -18,6 +18,7 @@ module SitePrism
     end
   end
 
+  class TimeOutWaitingForExistenceError < StandardError; end
   class TimeOutWaitingForElementVisibility < StandardError; end
   class TimeOutWaitingForElementInvisibility < StandardError; end
 
--- a/features/section.feature
+++ b/features/section.feature
@@ -60,6 +60,17 @@ Feature: Page Sections
     And I wait for the section element that takes a while to appear
     Then the slow section appears
 
+  Scenario: Wait for section - Exceptions - Positive
+    Given exceptions are configured to raise on wait_fors
+    When I navigate to the section experiments page
+    And I wait for the section element that takes a while to appear
+    Then the slow section appears
+
+  Scenario: Wait for section - Exceptions - Negative
+    Given exceptions are configured to raise on wait_fors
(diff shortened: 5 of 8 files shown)


ricmatsui:
@luke-hill Sounds good, for context in my current project I am debating between using the workaround (`raise WaitError unless wait_for_menu 10`) or monkey-patching to remove `wait_for_*` altogether and forcing usage of `menu wait: 10` to avoid errors not being raised when expected.

Additionally, I would like to also note that `wait_for_*` is the only method which does not raise errors (besides `has_*?/has_no_*?` which clearly should not) so I think it is a common pitfall or source of confusion. Perhaps we should add a note about this to the documentation in the meantime?

|Method|Error|
|--|--|
|"element/elements" | ExpectationNotMet|
|"section/sections"| ExpectationNotMet|
|"iframe"| NoSuchFrameError |
|wait_for_*| No error |
|wait_until_*_visible| TimeOutWaitingForElementVisibility|
|wait_until_*_invisible | TimeOutWaitingForElementInvisibility|

I'm not sure if you're aware, but there is also an experimental fork of this project by @twalpole that was just created which includes `wait_for_*` errors:

https://github.com/twalpole/capybara_page/blob/master/lib/capybara_page/element_container.rb#L130-L139

So I think there is a general desire to standardize the error behavior, which I think is part of the plan you have here as well.

tgaff:
It does indeed look like this is a hole in the API.

@ricmatsui Before I read through it all could you please rebase this on master?  It looks like there are a bunch of commits from Luke Hill on here that likely aren't supposed to be.

I did take a quick look and I'm curious why you went with the pattern you did, rather than what was done for .create_visibility_waiter and .create_invisibility_waiter


this change:
```rb
          result = Capybara.using_wait_time timeout do
            element_exists?(*find_args, *runtime_args)
          end
          raise SitePrism::TimeOutWaitingForExistence if !result && SitePrism.raise_on_wait_fors
          result
```

vs. 

.create_visibility_waiter and .create_invisibility_waiter
```rb
           Timeout.timeout timeout, SitePrism::TimeOutWaitingForElementVisibility do
             Capybara.using_wait_time 0 do
               sleep 0.05 until element_exists?(*find_args, *runtime_args, visible: true)
             end
           end
```
> Note This Capybara.using_wait_time inside of Timeout.timeout strikes me as a little odd.  Any insight here?


Also did we consider adding bang methods instead?

twalpole:
@tgaff using `Timeout.timeout` around any code that does anything complicated (network access, etc) is a terrible idea since the way it works can leave things in a completely unrecoverable state.

tgaff:
@twalpole Thanks for the info!  To be clear, I wasn't advocating Timeout.timeout.  I would advocate consistency though and I now see 3-4 different timeout implementations in SitePrism. 
 
I think SitePrism itself should try to avoid tinkering with timeouts as much as possible and leverage Capybara for that.  I know a huge amount of effort went into Capybara on that front and we should see how we can best use that.  Some of the early SitePrism code (admittedly several years ago) seemed to be at odds with Capybara and re-implemented functionality that Capybara devs had spent a *lot* of effort trying to get right.

twalpole:
@tgaff Yes - SitePrism does have the `use_implicit_waits` option which configures it to use Capybaras waiting but it defaults to false.   I believe when SitePrism started it was anti Capybaras waiting behavior for some reason, but then IMHO that just ends up leaving the user having to deal with timing issues and makes writing tests harder.

ricmatsuitile:
@tgaff I have rebased now (seems like master was reverted?)

> I did take a quick look and I'm curious why you went with the pattern you did, rather than what was done for .create_visibility_waiter and .create_invisibility_waiter

Mainly because of what you mentioned with using `Timeout.timeout` and `Capybara.using_wait_time` at the same time seemed odd, and what @twalpole mentioned with `Timeout` leaving things in an unrecoverable state. I have seen a test using wait for visible never time out (it ran continuously until I aborted manually) but I wasn't able to pinpoint the root cause yet since the repro rate is very low, but I suspect it is related to `Timeout.timeout` (I use `use_implicit_waits = true`)

I also did not want to introduce any additional changes in behavior which could be breaking for `wait_for` so I opted for this route which allows it to return the value using the same behavior as before (unless the raise flag is set)

> Also did we consider adding bang methods instead?

This is interesting approach, but that would still require a flag to opt into the behavior correct? For example the existing `wait_until_*_visible` would still raise an error if the opt in flag is not set, and *not* raise an error if the flag *is* set

It may also be more confusing with bang since it's not really modifying the page/section/element which I think is the convention for bang methods?

tgaff:
Thanks for the reply @ricmatsuitile (you know you switched user names?)

>  (seems like master was reverted?)
Err I hope master wasn't reverted, but my quick scan did not reveal where those commit messages had come from.  So err... ❔ 

Thank you for your explanation.  I think the way you're doing it is better.  Maybe I'll volunteer to cleanup the other waiting code.  It's at the very least not dry and as pointed out potentially dangerous (maybe less so post-1.9.3).  

### re: bang methods
I suppose you're right - that in order to migrate users toward it you'd need a flag.  *but see API normalization below*

In terms of the convention...it depends.  One way to sum up the convention might be to say that `!` methods symbolize 'danger' or potential 'surprise' ~ but don't tell you what kind of danger that is.  Rails presents the most obvious example of similar behavior on ActiveRecord::Base#save! and friends (which raise an error if a validation fails on record save while the non-bang method just returns false).  `Kernel.exit!` is a different example of a non-mutational bang method.

Bang methods would give users the choice to opt into the behavior anywhere without opting in everywhere.  That might be useful for some users.  Thoughts anyone?
*(I'm personally of the opinion that if `use_implicit_waits = true` were the only setting we could have saved a lot of trouble long ago 😞)*

### API normalization

If we are going to use a global setting to enable errors on `wait_for_*` new users might be surprised that they have to globally opt-in for `wait_for_menu` but others (like `wait_until_menu_visible`) enable raising an error by default (and indeed do not permit disabling it!)  
Do we need to add a similar `SitePrism.raise_on_wait_untils` ?  
Can we just deprecate the current behavior (with appropriate versioning/warnings) and get all of these on the same default behavior?

luke-hill:
Not read all of the comments yet (Was busy Thurs / Fri)
@twalpole - Thanks for all your input. I've seen you've tried to get an experimental branch up for Capybara 3 (Something I'd be happy to host on here if you want?) - Note we are considering slowly upping the minimum ruby requirements on this project, just I've only been the co-owner here for a short time so want to do things slow and clean up things first.

@tgaff Thanks for the offer for that work, please feel free to dive in. I'm still trying to establish the list of "What" needs doing first.

@ricmatsuitile - This now seems ok (Commit wise), I'll get round to reviewing this in the next day or so.

twalpole:
@luke-hill The branch I have is purely for experiments at the moment, making it Capybara 3 only, not including capybara/dsl into everything, removing the parts of site_prism I don't think make sense, etc.  so I don't think it makes sense to host it here at this point.

features/step_definitions/wait_steps.rb
```diff
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+When(/^it is configured to raise exceptions for wait fors$/) do
```
luke-hill: Lets make this a Given step and re-order the cukes accordingly.
`Given exceptions are configured to raise on wait_fors`

features/step_definitions/wait_steps.rb
```diff
@@ -12,6 +16,11 @@
   @test_site.home.wait_for_some_slow_element(2)
 end
 
+Then(/^an exception is raised if I wait for a specifically short amount of time for an element to appear$/) do
```
luke-hill: `... is raised when I wait for an element that won't appear`

lib/site_prism/element_container.rb
```diff
@@ -130,9 +130,11 @@ def create_waiter(element_name, *find_args)
       create_helper_method method_name, *find_args do
         define_method method_name do |timeout = nil, *runtime_args|
           timeout = timeout.nil? ? Waiter.default_wait_time : timeout
-          Capybara.using_wait_time timeout do
+          result = Capybara.using_wait_time timeout do
             element_exists?(*find_args, *runtime_args)
           end
+          raise SitePrism::TimeOutWaitingForExistence if !result && SitePrism.raise_on_wait_fors
```
luke-hill: as this is a new error, can we have them named with `Error` at the end. I'm aware the others don't but it's a breaking change to refactor them.

lib/site_prism/exceptions.rb
```diff
@@ -12,6 +12,7 @@ def message
 
   class NoSelectorForElement < StandardError; end
   class TimeoutException < StandardError; end
+  class TimeOutWaitingForExistence < StandardError; end
```
luke-hill: Again something I'm slowly trying to do is add in some nice error messages. If you look at `UnsupportedBlock` and design something similar that would be good.

ricmatsui: I decided to format the message at the point of raising, which I think is what is more standard to do in this case, let me know how I should be formatting the message here. I'm not familiar how to do it here when I want to format two arguments. I was having trouble setting the message when passing arguments to the error constructor.

ricmatsui:
Thanks @luke-hill @tgaff, I have completed the changes now, let me know if there are any more important items I should have tested. Apologies for delay I have been busy *using* site_prism 🙂

luke-hill:
Haha don't apologise for the delay. There is no rush in Open Source.

I'll probably not get round to reviewing this one for a week or so. Got some PR's of my own I need to work on a bit.

With regards to error message formatting, I'm not fussed at the moment about this as we are having a mini-debate with that.

README.md
```diff
@@ -1642,6 +1642,37 @@ with this:
...
 
+## Raising Errors on wait_for
+
+By default, when using `wait_for_*` methods, SitePrism will not raise
+an error if an element does not appear within the timeout period and will
+simply return `false` and allow the test to continue. This is different from
+the other methods such as `wait_until_*_visible` which do raise errors.
+
+Add the following code your spec_helper file to enable errors to be
+raised immediately when a `wait_for_*` does not find the element it is
```
luke-hill: a `wait_for_*` method

ricmatsui:
@luke-hill Thanks, I have updated this now. This also now turns on raising errors for the new `wait_for_no*` methods so they are also consistent.

lib/site_prism/element_container.rb
```diff
@@ -71,6 +71,20 @@ def raise_if_block(obj, name, has_block)
       raise SitePrism::UnsupportedBlock, "#{obj.class}##{name}"
     end
 
+    def raise_wait_for_if_failed(obj, name, timeout, failed)
+      return unless SitePrism.raise_on_wait_fors && failed
+
+      raise SitePrism::TimeOutWaitingForExistenceError, \
+            "Timed out after #{timeout}s waiting for #{obj.class}##{name}"
```
tgaff: These messages are very descriptive!  
But I feel like they're perhaps unnecessary and just one more thing to maintain (part of the reason why it got turned into its own method).  
The stacktrace already gives you the exact test code line the error occurs on and because of the design of SitePrism - it's already descriptive: 
`@test_site.home.wait_for_some_slow_element(5)` tells me everything in this message.

lib/site_prism/element_container.rb
```diff
@@ -71,6 +71,20 @@ def raise_if_block(obj, name, has_block)
       raise SitePrism::UnsupportedBlock, "#{obj.class}##{name}"
     end
 
+    def raise_wait_for_if_failed(obj, name, timeout, failed)
```
tgaff: These methods become part of the public API.

natritmeyer:
Some history:

Back when I started siteprism, capybara’s wait logic still wasn’t great. I asked around to see if it was just me who found it flakey; turns out everyone I asked had the same issues I was having. But, with time it improved radically.

lib/site_prism/element_container.rb
luke-hill: Whilst I agree with your premise. He's copied the existing structure.

This entire area needs some rework. I'd rather get this PR in as is (I've got some additional Items I'm going to be adding to the TODO document). If you're ok with merging this with this caveat I'll merge this in. I'd rather have you're agreement though.

lib/site_prism/element_container.rb
ricmatsui: Thanks, I mainly included this to match the behavior of the rest of the gem (timeout message and unsupported block message) and because I think it can be helpful for productivity in some cases, for example when the errors in Jenkins with a JUnit formatter, the Test Results page will show the message of the error and someone doesn't need to read the stack trace, but I agree the stacktrace contains the class and element information.

Do you think I should just include the timeout duration information?

luke-hill:
Merged. Thankyou for your contribution @ricmatsui

PR #229, opened 2018-03-22
SitePrism::Waiter.wait_until_true - Refactor

luke-hill:
Add new deprecation warning and tidy up class / specs


luke-hill:
ping @tgaff @TheMetalCode - Minor rubocop issue from last week now fixed. Ready for Review

lib/site_prism/waiter.rb
```diff
@@ -2,18 +2,22 @@
...
         sleep(0.05)
       end
-      raise SitePrism::TimeoutException, 'Timed out while waiting for block to return true'
+
+      raise SitePrism::TimeoutException, wait_time
     end
 
     def self.default_wait_time
-      Capybara.respond_to?(:default_max_wait_time) ? Capybara.default_max_wait_time : Capybara.default_wait_time
+      warn 'default_wait_time is now deprecated.'
```
tmertens: If there is a plan to remove it in a specific version, that should be specified in the message

lib/site_prism/exceptions.rb
```diff
@@ -11,7 +11,13 @@ def message
   end
 
   class NoSelectorForElement < StandardError; end
-  class TimeoutException < StandardError; end
+
+  class TimeoutException < StandardError
+    def message
+      "Timed out after #{super}s while waiting for block to evaluate as true."
```
tmertens: I don't think overriding `message` here is necessary or helpful, it would be better (IMO) to leave the message in the waiter class and just interpolate the time there. If I want to change the message in the future now I have to change two files instead of one

lib/site_prism/waiter.rb
luke-hill: For now there isn't. But I didn't want to just start deleting items in the codebase as Nat asked me to go easy on making sweeping changes.

We now have a restriction in terms of gems which negates the need for this method. But I'm leaving it in for now incase anyone still references it. I'll add in a note about it being removed shortly.

lib/site_prism/exceptions.rb
luke-hill: For variable messages I'd tend to agree with you there. However this message (And one or two others), I've opted to hardcode and standardize - With a view to getting more of the suite under a generic style / feel.

This error message only triggers in one place (and is rescued in another), and this occurs only when there is a loop that times out. As such I wouldn't want this message to be altered, or mutable.

luke-hill:
@tmertens / @TheMetalCode What are your thoughts on the PR as is now (Couple of minor updates done)?

PR #231, opened 2018-03-31
[wip] standardize waits

tgaff:
This is based from #226 and as such depends on that being merged first (at which point I'll rebase).  
For that reason I'm going to open this mainly for CI and then close it.

This attempts to remove the usage of Timeout.timeout as discussed in #226 and to standardize our use of waiting to use Capybara when possible.  (The Capybara team has put a lot of effort into that aspect of Capybara and we should leverage it whenever possible.)  As such I added a new waiter method `wait_with_capybara` on the utility `Waiter` class.  This method optionally takes an Error that will be raised if Capybara fails in the block passed to it.  (I did consider putting this on Capybara itself but this is probably the safer approach.)

Also, I wrote this on a flight to Korea where I'll be on vacation for the next few weeks - so just FYI I might be pretty quiet for a bit, sorry!  (mostly @ Luke)


lib/site_prism/waiter.rb
```diff
@@ -12,6 +12,11 @@ def self.wait_until_true(wait_time_seconds = default_wait_time)
       raise SitePrism::TimeoutException, 'Timed out while waiting for block to return true'
     end
 
+    def self.wait_with_capybara(timeout = self.default_wait_time, or_raise: nil)
+      raise or_raise unless (result = Capybara.using_wait_time(timeout) { yield }) || !or_raise
+      result
+    end
+
```
tgaff: Considering renaming `or_raise` to something better...maybe `raise_on_fail`?  Suggestions?

README.md
```diff
@@ -630,7 +630,7 @@ it wil return `false`. For example, with the following page:
 
 ```ruby
 class Friends < SitePrism::Page
-  elements(:names, 'ul#names li a')
+  elements :names, 'ul#names li a'
```
tgaff: The rest of the doc used this format - standardizing.

tgaff:
OK, CI checks out.  I'll close this until 226 is picked up.

PR #234, opened 2018-04-10
Cucumber tidy

luke-hill:
I have tried to make the individual scenarios split out better into their relevant areas. Done a mini-bit of DRYing and also tried to add some coverage where it was missing (Note the changes to the sample pages)

I've also tried to (Where I've spotted it), introduce some naming conventions that indicate what type of method / test is being performed.


luke-hill:
Ping @TheMetalCode for thoughts (Mainly refactoring though)

PR #235, opened 2018-04-11
Minor developmental update

luke-hill:
- Allow later versions of selenium on local
- Readme minor style tweaks
- Cleanup spec_helper and alter default formatter to documentation
- Remove default JS driver to be selenium (Already set)


luke-hill:
ping @TheMetalCode

spec/spec_helper.rb
```diff
@@ -21,20 +20,12 @@
 require 'pages/home'
 
 RSpec.configure do |config|
-  config.expect_with :rspec do |c|
```
TheMetalCode: Curious why we wouldn't just explicitly force `expect` here? I agree that this block as it stood was superfluous since both `expect` and `should` are allowed by default. But the RSpec docs all but beg you to just use expect. https://relishapp.com/rspec/rspec-expectations/docs/syntax-configuration

luke-hill: If you read the Pseudo-cuke below it which shows the first use case (No configuration)

```
Given a file named "spec/spec_helper.rb" with: # Empty file
When I run rspec
Then the examples should all pass
And the output should contain "Using should from rspec-expectations' old :should syntax without explicitly enabling the syntax is deprecated"
```

You'll see that both commands work. But `should` throws a warning. We could force `expect` but in my eyes this feels a trivial use case (We now have RSpec 3+ which throws these warnings, we previously didn't), we have community code reviews (Again previously didn't), and on the whole not many people would code with this style.

We could put in the restriction. But anyone who codes with `should` will be caught out by me at CR. I'd also be interested to note the number of people who still do this.

TheMetalCode: RSpec configs don't govern cucumber feature syntax though, but only the specific RSpec assertion syntax in either the unit tests or the cucumber step definitions. So the pseudo cuke cited would be unaffected.

`it` blocks should also be unaffected, so `it 'should do something'` would also be fine.

Near as I can tell on latest master, `should` does not appear at all in `features/step_definitions`, and in `spec` I only see `should` in it blocks and not any actual assertions. So forcing `expect` assertions really should be safe. I tested this out locally just to be sure.

luke-hill: I'm not quite sure what you're referring to there. But I think you're in agreement with me. So I'll just merge this in.

I know RSpec configs only cover the rspec portions of the codebase (Unit tests e.t.c.). And as you've mentioned about "should" never being used this is fine.

PR #238, opened 2018-04-16
Bump cucumber to 3.0.1

luke-hill:
Note that this will involve a bump to our minimum ruby version requirements (Which has until now been locked at 2.0). I have opted to pick the latest version of cucumber which can be ran at Ruby 2.1, because at least this isn't as big of an upgrade.

However I don't see this as an issue given we need to be running our integration tests against Cucumber3 (It is slightly quicker). And Ruby 2.0 has been EOL for ages.

Also down the line other gem requirements have higher version dependencies, so this is in-line with the other ones.

### DIFF
--- a/features/step_definitions/iframe_steps.rb
+++ b/features/step_definitions/iframe_steps.rb
@@ -1,54 +1,54 @@
 # frozen_string_literal: true
 
-Then(/^I can locate the iframe by id$/) do
+Then('I can locate the iframe by id') do
   @test_site.home.wait_for_my_iframe
 
   expect(@test_site.home).to have_my_iframe
 end
 
-Then(/^I can locate the iframe by index$/) do
+Then('I can locate the iframe by index') do
   @test_site.home.wait_for_index_iframe
 
   expect(@test_site.home).to have_index_iframe
 end
 
-Then(/^I can locate the iframe by name$/) do
+Then('I can locate the iframe by name') do
   @test_site.home.wait_for_named_iframe
 
   expect(@test_site.home).to have_named_iframe
 end
 
-Then(/^I can locate the iframe by xpath$/) do
+Then('I can locate the iframe by xpath') do
   @test_site.home.wait_for_xpath_iframe
 
   expect(@test_site.home).to have_xpath_iframe
 end
 
-Then(/^I can see elements in an iframe$/) do
+Then('I can see elements in an iframe') do
   @test_site.home.my_iframe do |f|
     expect(f.some_text.text).to eq('Some text in an iframe')
   end
 end
 
-Then(/^I can see elements in an indexed iframe$/) do
+Then('I can see elements in an indexed iframe') do
   @test_site.home.index_iframe do |f|
     expect(f.some_text.text).to eq('Some text in an iframe')
   end
 end
 
-Then(/^I can see elements in a named iframe$/) do
+Then('I can see elements in a named iframe') do
   @test_site.home.named_iframe do |f|
     expect(f.some_text.text).to eq('Some text in an iframe')
   end
 end
 
-Then(/^I can see elements in an xpath iframe$/) do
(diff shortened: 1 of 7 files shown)


luke-hill:
Ping @TheMetalCode @tgaff - Ignore CI failures. It's only the 2.0 branch failing which will be removed when I rebase this with other work.

luke-hill:
Yes it would. Good spot. 

On a side note do you think this should be considered for 2.13 (As I've initially intimated), which would be released in May, or would you perhaps think it's more prudent to target 2.14 (Which would probably be June / July)

crcraig92:
I've been poking around on here to figure out why the dependency for site_prism has capybara locked and doesn't allow 3.0.2.  If you let capybara pick a version of site_prism it reverts all the way back to 1.3.

This seems to be the most related post I've stumbled upon.

Is this a blocker for supporting the newer versions of capybara?

I'm running into issues with Firefox and some of them seem like they may be being fixed in capybara and I'd like to test newer versions before filing bug reports.

tgaff:
@crcraig92 This shouldn't be blocking that - the cucumber change is only a **development** dependency.  
The gemspec does lock Capybara to 2.x series though.

crcraig92:
@tgaff Thanks. I didn't think this would be the case but I wanted to be sure before I went creating new issues.

tgaff:
@crcraig92 If you want, feel free to create an issue to support Capybara 3.x; there's not one open and it may help to track interest and focus work on that.

tgaff:
@luke-hill 
Is updating cucumber worth dropping support for ruby 2.0?  (I'm not a cucumber user.)

If so and if there are current users stuck on ruby 2.0 then it might be nice to get them as many of the current slated updates (in 2.13) as possible first.  However, I don't see any that are super beneficial to users so if you want this in 2.13 I won't object.

### From what I see 2.13 currently includes:
* SitePrism::Waiter.wait_until_true - Refactor #229
* Doc updates #240 
* Project Ruby Settings #237

I wish I had info on whether there are actually any gem users stuck on ruby 2.0!

crcraig92:
@tgaff I have done this
[Support Capybara 3.x](https://github.com/natritmeyer/site_prism/issues/242)

Thanks again for your input!

luke-hill:
@tgaff It's a valid comment again. You keep doing this! Honestly. I'm not 100% sure. I know at my current company we will go to cucumber3 once jetbrains support it.

I think at runtime cucumber3 is a bit better (But again this is only a dev dependency for us, so perhaps we shouldn't).

See comments on @crcraig92 thread for other items.

Ping @TheMetalCode @RustyNail - Any thoughts on this?

RustyNail:
I think that these versions should be consistent with each other.
- `required_ruby_version` defined in .gemspec (2.1.0)
- .ruby-version (2.2.0)
- `TargetRubyVersion` defined in .rubocop.yml (2.2.0)

It also exists as a verification item of rubocop.
- http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Gemspec/RequiredRubyVersion

And it is recommended to update to the new version because ruby 2.1 series has finished the support period.

luke-hill:
@RustyNail if you want to help with the maintenance of site_prism feel free to tackle more of the items in the TODO document in the repo. Or if you feel there are other areas to contribute go for it.

Having the 3 ruby versions you mentioned in sync is correct. However what this allows us to do (By not having them in sync). Is have unsupported versions of ruby (Which still work).

This means we can say we only support Say Ruby 2.1+, but if you use Ruby 2.0, and you can fix your own problems, then we won't inhibit your usage.

However going forwards. Combined with the fact Ruby 2.0 is so EOL it's unreal. Plus the fact we want to start supporting later capybara versions. I think updating the Minimum ruby requirement for this gem is going to happen. So I think this should be done here first. (Small update)

PR #241, opened 2018-04-20
Minor dev tweaks

luke-hill:
Fix some syntax
Alter rspec formatter to documentation
Allow Selenium 3.9 and 3.10 to be used for testing

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,35 +66,35 @@ end
 
 # now for some tests
 
-When /^I navigate to the google home page$/ do
+When(/^I navigate to the google home page$/) do
   @home = Home.new
   @home.load
 end
 
-Then /^the home page should contain the menu and the search form$/ do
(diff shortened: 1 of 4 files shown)


luke-hill:
Ping @tgaff @TheMetalCode @tmertens (If you're interested in reviewing stuff)

README.md
```diff
@@ -66,35 +66,35 @@ end
 
 # now for some tests
 
-When /^I navigate to the google home page$/ do
+When(/^I navigate to the google home page$/) do
```
tgaff: I'm not a big cucumber user, but I personally find the parentheses-less format nicer, more ruby-ish and more readable.
I do however like the use of strings when the regex is not necessary.

luke-hill: Fair point. When we migrate to cucumber 3 (at whatever point), the regex style (for us), will become unrequired. As we don't have capture groups, and can use direct strings.

PR #248, opened 2018-05-04
Fix weird waiting bug - Hard to triage

luke-hill:
So in essence, this PR does 2 things. It fixes this bug, but then also removes the previously deprecated code.

For some reason (Hard to determine why, and if it was affecting anyone in production), `Waiter.default_wait_time` seems to be calling the Capybara method, instead of the class method defined this repo. So what I did to remove the affliction was to rename the method to `sp_default_wait_time` and all associated calls. This then fixed the issue, however then our Spec run times shot up.

So this required fixing the `spec_helper.rb` to run with the wait times set to 0 (So the specs now run fast again). However after doing this, I noticed that the warning message was now incorrect (As such this PR will hit in `2.14` not `2.13`)

Once I fixed all that up. I noticed that because I've fixed all the calls, and the developmental / testing side works again as expected, the method had no inter-suite calls. So I've removed it, and tidied up the suite.

This now fixes the bug we were seeing, doesn't cause any adverse affects, and the only noticeable change is that we now don't have a 2nd class method in the Waiter class (Which shouldn't be used publically anyway).

### DIFF
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -148,8 +148,7 @@ module SitePrism
     def create_waiter(element_name, *find_args)
       method_name = "wait_for_#{element_name}"
       create_helper_method(method_name, *find_args) do
-        define_method(method_name) do |timeout = nil, *runtime_args|
-          timeout = timeout.nil? ? Capybara.default_max_wait_time : timeout
+        define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
           result = Capybara.using_wait_time(timeout) do
             element_exists?(*find_args, *runtime_args)
           end
@@ -162,8 +161,7 @@ module SitePrism
     def create_nonexistence_waiter(element_name, *find_args)
       method_name = "wait_for_no_#{element_name}"
       create_helper_method(method_name, *find_args) do
-        define_method(method_name) do |timeout = nil, *runtime_args|
-          timeout = timeout.nil? ? Waiter.default_wait_time : timeout
+        define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
           result = Capybara.using_wait_time(timeout) do
             element_does_not_exist?(*find_args, *runtime_args)
           end
--- a/development_setup.md
+++ b/development_setup.md
@@ -22,4 +22,3 @@ Happy Testing / Developing!
 
 Cheers,
 The SitePrism Team
-
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -40,7 +40,8 @@ module SitePrism
 
     def displayed?(*args)
       expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
-      seconds = !args.empty? ? args.first : Waiter.default_wait_time
+      seconds = !args.empty? ? args.first : Capybara.default_max_wait_time
+
       raise SitePrism::NoUrlMatcherForPage if url_matcher.nil?
       begin
         Waiter.wait_until_true(seconds) { url_matches?(expected_mappings) }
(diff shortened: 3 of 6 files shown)


luke-hill:
Ping @tgaff @TheMetalCode @RustyNail - Thoughts / review points.

lib/site_prism/element_container.rb
```diff
@@ -147,7 +147,7 @@ def create_nonexistence_waiter(element_name, *find_args)
       method_name = "wait_for_no_#{element_name}"
       create_helper_method(method_name, *find_args) do
         define_method(method_name) do |timeout = nil, *runtime_args|
-          timeout = timeout.nil? ? Waiter.default_wait_time : timeout
+          timeout = timeout.nil? ? Capybara.default_max_wait_time : timeout
```
RustyNail: How about writing like this?

```
define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
```

The `create_visibility_waiter` method just below it is written as such.

luke-hill: Indeed it is, not sure why. Will standardise up

tgaff:
@luke-hill I just saw this.  Do you have a test case we can run against #251 which refactors all of that?

luke-hill:
@tgaff Once this is merged in, we can test against it. But this will not hit the current RC. (I'll probably try cut something soon as we've got enough changes for it to be significant)

lib/site_prism/element_container.rb
```diff
@@ -148,8 +148,7 @@ def create_nonexistence_checker(element_name, *find_args)
     def create_waiter(element_name, *find_args)
       method_name = "wait_for_#{element_name}"
       create_helper_method(method_name, *find_args) do
-        define_method(method_name) do |timeout = nil, *runtime_args|
-          timeout = timeout.nil? ? Capybara.default_max_wait_time : timeout
+        define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
```
ineverov: The line is too long. Pls, make it 80.

lib/site_prism/element_container.rb
```diff
@@ -162,8 +161,7 @@ def create_waiter(element_name, *find_args)
     def create_nonexistence_waiter(element_name, *find_args)
       method_name = "wait_for_no_#{element_name}"
       create_helper_method(method_name, *find_args) do
-        define_method(method_name) do |timeout = nil, *runtime_args|
-          timeout = timeout.nil? ? Waiter.default_wait_time : timeout
+        define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
```
ineverov: The line is too long. Pls, make it 80.

tgaff:
@luke-hill  I'm a little hazy on the cause of the issue here, but the reason I asked was that 251 will alter the same logic as what you're doing - effectively replacing the changes here.  I was hoping that you could give me steps to reproduce this issue or even better a spec that demonstrates it so that I can verify that 251 does not re-introduce it.  Thanks!

luke-hill:
Calling the deprecated method doesn't actually occur in the stack because of a name conflict when called inter-method.

Changing the name forces the call.

This change ensures that the now removed method will be called (Even after removal, by just calling its method signature)

lib/site_prism/element_container.rb
luke-hill: There are some tasks in the todo document to reduce some of the cop infractions. Currently the max line length is set to something like 130.

It needs reducing I agree, but not here.

ineverov:
@luke-hill sorry, I'm still not able to reproduce it in the original code... I don't see method named `timeout` either.
Could you, please, provide steps to reproduce?

luke-hill:
@tgaff Once this is merged in and you've fixed up 251 we'll be fine. Essentially so long as we don't have the deprecated Waiter class method we'll be fine.

PR #249, opened 2018-05-04
Final cucumber cleanup

luke-hill:
I've moved the steps around a bit, and done a couple of renames. But frankly I want to leave this as is for a while and see what others think. And also move onto some actual feature work / bug fixes


luke-hill:
ping @tgaff - Happy for this refactor to go in (Will probably cut the release after this)

Issue #85, opened 2014-07-13
Provide ability to set locator of root element in section class


abotalov:
Consider the following example (modified from SitePrism README):

``` ruby
class MenuSection < SitePrism::Section
end
class Home < SitePrism::Page
  section :menu, '#gbx3', MenuSection
end
class SearchResults < SitePrism::Page
  section :menu, '#gbx3', MenuSection
end
```

So `#gbx3` should currently be specified in several places which isn't DRY.

IMO in many situations a section (i.e. menu/search bar/authentication form) is likely to have the same locator of root element that is specified in partial.

Therefore, I propose to add ability to do the following:

``` ruby
class Menu < SitePrism::Section
  set_root_locator '#menu'
end
class Home < SitePrism::Page
  section :menu, Menu
end
class SearchResults < SitePrism::Page
  section :menu, Menu
end
```

For example HTMLElements seem to also provide such ability - https://github.com/yandex-qatools/htmlelements#create-blocks-of-elements (using `@Block(@FindBy(xpath = "//form"))` annotation

If it will be approved I can do PR.

Tensho:
+1

tgaff:
:+1: I also wanted something like this over a year ago.

tgaff:
btw, @abotalov I think your code example for section should say:
`  section :menu, MenuSection, "#gbx3"`

luke-hill:
Hi @abotalov This dovetails with some other items around similar areas that people have mentioned. I think you've hit something though that feels a bit safer than the other topics mentioned.

Was this more of a brainstorm / idea, or something that you had mocked up. What are your thoughts with this going forwards? To land a PR or to keep this active for more discussion. Or just to close?

tgaff:
I agree with what @luke-hill says above.  This PR and #161 are both affecting similar code and how `section` is to be called.  

Starting with the original:
`section :menu, MenuSection, "#gbx3"`

PR #161  proposes to find the section's class automatically (via convention) resulting in:
`section :menu, "#gbx3"  `

This PR #85 proposes 
`section :menu, Menu`


Combining the two would yield the clean and wonderfully simple:
`section :menu`

Which to me at least seems quite nice and I think would cover many section use cases.  

(1)
We might want to think about whether these changes should come at the same time or not.  I think for the simplicity of anyone using this, doing them simultaneously will allow us to be as clear as possible and reduce the amount of potentially confusing cross-messaging if they were to strike a couple of weeks apart.  

(2)
[...]

PR #250, opened 2018-05-04
[feat] Allow to set search arguments inside section class

ineverov:
Issue #85

## Changes
* Allow to provide default search arguments inside section definition
  *  If section has its own search arguments, default search arguments will be ignored
  * If section has no search arguments section class or any parent class deafult arguments will be used
* Allow to provide section class together with a block
* Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -804,11 +804,43 @@ SitePrism allows adding sections to sections) is to call the `section`
 method. It takes 3 arguments: the first is the name of the section as
 referred to on the page (sections that appear on multiple pages can be
 named differently). The second argument is the class of which an
-instance will be created to represent the page section, and the third
-argument is a css selector that identifies the root node of the section
+instance will be created to represent the page section, and the following
+arguments are [Capybara::Node::Finders](https://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Finders).
+These identify the root node of the section
 on this page (note that the css selector can be different for different
 pages as the whole point of sections is that they can appear in
 different places on different pages).
+
+If you define a section as a class and as an Anonymous section,
+you can have some handy constructs like the one below
+
+```ruby
+class People < SitePrism::Section
+  element :headline, 'h2'
+end
+
+class HomePage < SitePrism::Page
+  # section people_with_block will have headline and
+  # headline_clone elements in it
+  section :people_with_block, People do
+    element :headline_clone, 'h2'
+  end
+end
+```
+
+The 3rd argument (Locators), can be omitted if you are re-using the same locator for all
+references to the section Class. In order to do this, simply tell SitePrism that
+you want to use a default search argument.
+
+```ruby
+class People < SitePrism::Section
+  set_default_search_arguments '.people'
+end
+
+class Home < SitePrism::Page
+  section :people, People
+end
+```
 
 #### Accessing a page's section
 
--- a/features/section.feature
+++ b/features/section.feature
@@ -22,6 +22,11 @@ Feature: Page Sections
     When I navigate to the section experiments page
     Then I can see a section within a section using nested blocks
 
+  Scenario: section within a section using class & blocks
+    When I navigate to the home page
+    Then I can see elements from the parent section
+    And I can see elements from the block
+
   Scenario: section scoping
     When I navigate to the home page
     Then access to elements is constrained to those within the section
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -227,13 +227,36 @@ module SitePrism
 
     def extract_section_options(args, &block)
       if args.first.is_a?(Class)
-        section_class = args.shift
-      elsif block_given?
-        section_class = Class.new(SitePrism::Section, &block)
-      else
-        raise ArgumentError, 'You should provide section class either as a block, or as the second argument.'
+        klass = args.shift
+        section_class = klass if klass.ancestors.include?(SitePrism::Section)
       end
-      [section_class, args]
+
+      section_class = deduce_section_class(section_class, &block)
+      arguments = deduce_search_arguments(section_class, args)
+      [section_class, arguments]
+    end
+
+    def deduce_section_class(base_class, &block)
+      klass = base_class
+
+      klass = Class.new(klass || SitePrism::Section, &block) if block_given?
+
+      unless klass
+        raise ArgumentError, "You should provide descendant of SitePrism::Section \
+class or/and a block as the second argument."
+      end
+      klass
+    end
+
+    def deduce_search_arguments(section_class, args)
+      extract_search_arguments(args) ||
+        extract_search_arguments(section_class.default_search_arguments) ||
+        raise(ArgumentError, "You should provide search arguments \
+in section creation or set_default_search_arguments within section class")
+    end
+
+    def extract_search_arguments(args)
+      args if args && !args.empty?
     end
   end
 end
--- a/lib/site_prism/section.rb
+++ b/lib/site_prism/section.rb
@@ -11,6 +11,19 @@ module SitePrism
 
     attr_reader :root_element, :parent
 
+    def self.set_default_search_arguments(*args)
+      @default_search_arguments = args
+    end
+
+    def self.default_search_arguments
+      @default_search_arguments ||
+        (
+          superclass.respond_to?(:default_search_arguments) &&
+          superclass.default_search_arguments
+        ) ||
+        nil
+    end
+
     def initialize(parent, root_element)
       @parent = parent
       @root_element = root_element
--- a/spec/section_spec.rb
+++ b/spec/section_spec.rb
@@ -7,6 +7,42 @@ class Page < SitePrism::Page; end
 
 describe SitePrism::Page do
   describe '.section' do
+    it 'should be callable' do
+      expect(SitePrism::Page).to respond_to(:section)
+    end
+
+    describe '.section with class and block' do
+      class SingleSection < SitePrism::Section
+        element :single_section_element, '.foo'
+      end
+
+      class PageWithSections < SitePrism::Page
+        section :single_section, SingleSection, '.bob'
+
+        section :section_with_a_block, SingleSection, '.bob' do
+          element :block_element, '.btn'
+        end
+      end
+
+      subject { PageWithSections.new }
+
+      before do
+        expect(subject).to receive(:find_first).and_return(:element)
+      end
+
+      it 'should be an instance of provided section class' do
+        expect(subject.section_with_a_block.class.ancestors).to include(SingleSection)
+      end
+
+      it 'should have elements from the base section' do
+        expect(subject.section_with_a_block).to respond_to(:single_section_element)
+      end
+
+      it 'should have elements from the block' do
+        expect(subject.section_with_a_block).to respond_to(:block_element)
+      end
+    end
+
     context 'second argument is a Class' do
       class PageWithSection < SitePrism::Page
         section :section, Section, '.section'
@@ -20,8 +56,21 @@ describe SitePrism::Page do
 
     context 'second argument is not a Class and a block given' do
       class PageWithAnonymousSection < SitePrism::Page
-        section :anonymous_section, '.section' do |s|
-          s.element :title, 'h1'
+        section :anonymous_section, '.section' do
+          element :title, 'h1'
+        end
+      end
+
+      subject { PageWithAnonymousSection.new }
+
+      it { is_expected.to respond_to(:anonymous_section) }
+      it { is_expected.to respond_to(:has_anonymous_section?) }
+    end
+
+    context 'second argument is a Class and a block given' do
+      class PageWithAnonymousSection < SitePrism::Page
+        section :anonymous_section, Section, '.section' do
+          element :title, 'h1'
         end
       end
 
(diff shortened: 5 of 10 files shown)


ineverov:
going to work on docs/feature tests later this or next week

luke-hill:
Option 1 definitely.

If you provide this argument in the Class, use that.
If you provide the argument in the definition DSL, use that.
If you provide the argument in the definition DSL AND the Class, use the definition DSL (No smart merging)

lib/site_prism/section.rb
```diff
@@ -11,6 +11,30 @@ class Section
 
     attr_reader :root_element, :parent
 
+    class << self
```
luke-hill: This area will probably need some explaining to me, as I'm not quite sure on the moving parts. Would you be willing to jump on slack?

spec/section_spec.rb
```diff
@@ -49,6 +118,38 @@ class PageWithAnonymousSection < SitePrism::Page
   let(:locator) { instance_double('Capybara::Node::Element') }
   let(:section_with_block) { SitePrism::Section.new(Page.new, locator) { 1 + 1 } }
 
+  describe '#default_search_arguments' do
+    let(:base_section) { Class.new(SitePrism::Section) }
```
luke-hill: I'd prefer a more legible integration-y style mock here. With a physical Page Object being instantiated through a more human interface.

Something like https://github.com/natritmeyer/site_prism/pull/250/files#diff-6fa85b9a8ee926d0d01e58459cddac2cR11 would be a good basis in my eyes

spec/section_spec.rb
```diff
@@ -49,6 +118,38 @@ class PageWithAnonymousSection < SitePrism::Page
...
+        section.set_default_search_arguments '//body', text: 'Body'
+      end
+
+      it 'should return default search arguments' do
+        expect(arguments).to eql(['//body', text: 'Body'])
+      end
+
+      context 'if base class default_search_arguments are set' do
+        before do
+          base_section.set_default_search_arguments :xpath
```
luke-hill: This seems erroneous, doesn't this just imply that you've set a base searching path of ... nothing? I know you're in a WIP so ignore this if irrelevant

lib/site_prism/section.rb
ineverov: yes, sure. Give me a link to it pls

spec/section_spec.rb
ineverov: I did it for the particular reason. Defining class integration-y will affect all unit tests and may potentially lead to false positive results. Because class will remain the same for all the test and redefinition/changes to this class will affect all the tests. For example,
```ruby
context '#hello' do
  class A
     def hello; 'Hello'; end
  end
  it 'should return hello' do
     expect(A.new.hello).to eql('Hello')
  end
end

context '#hi' do
  class A
     def hello; 'hi'; end
  end
  it { expect(A.new.hello).to eql('hi')}
end
```
in this case first example will fail

spec/section_spec.rb
ineverov: the idea was to allow to concatenate search arguments. something like
```ruby
#base class
set_default_search_arguments :xpath, '//html'

#child class
set_default_search_arguments text: 'Name'

# will produce search arguments like: 
# :xpath, '//html', text: 'Name'
```

But I'm going to change it as you recommended in PR comment to always overrride with child arguments if they present

ineverov:
@luke-hill @tgaff Review pls

lib/site_prism/section.rb
luke-hill: Happy for this and the below (Although not sure why you can't just use class methods). (Not happy with the memoized default_search_args (2 methods below), but mentioned on slack

spec/section_spec.rb
luke-hill: Correct, due to overloading. However I'm in favour of doing some named examples scoping and having these classes reinstantiated with differing layouts (xpath / css), for example. And have a chunky PR reworking a load of specs for this.

lib/site_prism/section.rb
ineverov: It's actually the same as using class methods. I can change it to `def self.` if you prefer this

spec/section_spec.rb
ineverov: Ok, I'll change it

lib/site_prism/section.rb
luke-hill: If it doesn't facilitate a big rewrite could you (#newbie)

test_site/pages/home.rb
```diff
@@ -30,7 +30,10 @@ class TestHomePage < SitePrism::Page
   element :nonexistent_element, 'input#nonexistent'
 
   # individual sections
-  section :people, People, '.people'
+  section :people, People
```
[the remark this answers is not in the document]
ineverov: It is used in some tests I guess, but yes, I can use below section and rename it to just people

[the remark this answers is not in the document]
luke-hill: Actually I've re-read the part above. And you're showing that actually both of these references are the same class object. I'm happy with that.

Could you add a comment to the first item. Saying that the locator has been defined elsewhere and each ref will have both element objects in it

spec/sections_spec.rb
```diff
@@ -22,6 +26,20 @@ class Page < SitePrism::Page
...
+  describe '.section with class and block' do
+    before do
+      expect(subject).to receive(:find_first).and_return(:element)
+    end
+
+    it 'should be an instance of provided section class' do
+      expect(subject.section_with_a_block.class.ancestors).to include(SingleSection)
+    end
+
+    it 'should have elements from the block' do
```
luke-hill: Should we therefore add an extra test `it 'has elements from the defined class'`

Then add an extra element definition after opening the class in this file on line8, call it whatever makes you happy. (Just add one)

.overcommit.yml
```diff
@@ -0,0 +1,31 @@
+# Use this file to configure the Overcommit hooks you wish to use. This will
```
luke-hill: Is this file meant to be in here>? If it's not, feel free to edit our gitignore

README.md
```diff
@@ -805,10 +805,52 @@ method. It takes 3 arguments: the first is the name of the section as
 referred to on the page (sections that appear on multiple pages can be
 named differently). The second argument is the class of which an
 instance will be created to represent the page section, and the third
-argument is a css selector that identifies the root node of the section
+argument the search argument that identifies the root node of the section
```
luke-hill: This doesn't quite read correctly, probably should be reverted

README.md
```diff
@@ -805,10 +805,52 @@ method. It takes 3 arguments: the first is the name of the section as
...
 on this page (note that the css selector can be different for different
 pages as the whole point of sections is that they can appear in
-different places on different pages).
+different places on different pages). It can also take the block with
```
luke-hill: This definition is defined under Anonymous sections, so isn't required.

README.md
```diff
@@ -805,10 +805,52 @@ method. It takes 3 arguments: the first is the name of the section as
...
+
+```ruby
+class SectionExperiments < SitePrism::Page
+  sections :level_1, '.level-1' do
+    element :element1, '.element-1'
+  end
+end
+```
+
+Providing class and block allows also to extend section for this
```
luke-hill: This can be rewritten to something like ...
```
If you define a section as a class and as an Anonymous section, you can have some handy constructs like the one below

<insert ruby code>
```

I'd also move this advice to below here: https://github.com/natritmeyer/site_prism/pull/250/files#diff-04c6e90faac2675aa89e2176d2eec7d8R1180

README.md
```diff
@@ -805,10 +805,52 @@ method. It takes 3 arguments: the first is the name of the section as
...
+  # section people_with_block will have headline and
+  # headline_clone elements in it
+  section :people_with_block, People do
+    element :headline_clone, 'h2'
+  end
+end
+```
+
+
+CSS selector can be ommited if you have same locator for all the sections
```
luke-hill:
```
The CSS selector can be omitted if you are re-using the same locator for all
references to the section Class. In order to do this, simply tell SitePrism that
you want to use a default search argument.

<ruby code>
```

EDIT: Delete one of the two lines above.

features/section.feature
```diff
@@ -22,6 +22,11 @@ Feature: Page Sections
     When I navigate to the section experiments page
     Then I can see a section within a section using nested blocks
 
+  Scenario: section within a section using class & blocks
+    When I navigate to the home page
+    Then I can see elements from the parent section
+    Then I can see elements from the block
```
luke-hill: `And`

lib/site_prism/element_container.rb
```diff
@@ -228,14 +228,37 @@ def deduce_iframe_element_find_args(args)
...
-        section_class = args.shift
-      elsif block_given?
-        section_class = Class.new(SitePrism::Section, &block)
+      section_class = deduce_section_class(args, &block)
+      arguments = deduce_search_arguments(section_class, args)
+      [section_class, arguments]
+    end
+
+    def deduce_section_class(args, &block)
+      if args.first.is_a?(Class) && args.first.ancestors.include?(SitePrism::Section)
```
luke-hill: Apologies if I'm missing something here, but this condition isn't always successful, even if a block is given.

This means that `base_class` can sometimes be an undefined local variable right?

lib/site_prism/element_container.rb
```diff
@@ -228,14 +228,37 @@ def deduce_iframe_element_find_args(args)
...
+
+      unless klass
+        raise ArgumentError, "You should provide section class either \
+as a block, or as the second argument."
+      end
+      klass
+    end
+
+    def deduce_search_arguments(section_class, args)
+      if args && !args.empty?
```
luke-hill: for both this call and the one below we've got a double chain just checking for their presence safety before evaluating their length.

Could we not maybe create some sort of private method like `present?` like I've done in another PR that takes an array input. Or is that over-analysing do you think?

lib/site_prism/element_container.rb
ineverov: yes, you are right, in this case, the interpreter will treat it as `nil` [on line 242](https://github.com/natritmeyer/site_prism/pull/250/files#diff-0142fe7beaa583d4ce8030ab2a152817R242) and it will use SitePrism::Section as a base class

README.md
ineverov: Sorry, English is not my native language, so I appreciate any suggestions how to improve wording

lib/site_prism/element_container.rb
```diff
@@ -228,14 +228,36 @@ def deduce_iframe_element_find_args(args)
...
+    end
+
+    def deduce_search_arguments(section_class, args)
+      extract_search_arguments(args) ||
+        extract_search_arguments(section_class.default_search_arguments) ||
+        raise(ArgumentError, "You should provide search arguments \
+in section creation or set_default_search_arguments within section class")
+    end
+
+    def extract_search_arguments(args)
```
ineverov: I guess, this method can be used in the future to do more work, i.e. converting site_prism's search arguments to capybara's (Now they are the same match, but who knows what they'll change in the future). IMO, it's not good to have dependencies on internal implementation (capybara dsl).

lib/site_prism/element_container.rb
luke-hill: Qu) When I do `var1 || 'string'` inside a ruby interpreter it crashes. Are you sure this will work? Aren't only instance variables `nil` by default? Or am I missing a further point?

spec/section_spec.rb
```diff
@@ -41,6 +41,57 @@ class PageWithAnonymousSection < SitePrism::Page
           .with_message(error_message)
       end
     end
+
+    context 'default search arguments' do
+      class PageWithSectionWithDefaultSearchArguments < SitePrism::Page
```
luke-hill: I've not fully read through these yet. But this format for tests I prefer a lot! Top work!

lib/site_prism/element_container.rb
ineverov: Sorry, I told you part of the truth...
```ruby
var1 || 'string' # Raises NameError
```

```ruby
if false # never can be reached
  var1 = 1
end

var1 || 'string' # Works. It's a part of how the ruby interpreter works. =(
```

spec/section_spec.rb
ineverov: 👍

luke-hill:
pin @tgaff @ricmatsui 

I need to have one final read through of this, but I've asked a few questions. If you guys have any comments can you add them in the next 7-10days?

spec/section_spec.rb
```diff
@@ -41,6 +41,57 @@ class PageWithAnonymousSection < SitePrism::Page
...
+        end
+      end
+
+      context 'if default search arguments are not set and no search arguments provided' do
+        it 'should raise ArgumentError' do
+          expect do
+            class ErroredPage < SitePrism::Page
+              section :section, Section
+            end
+          end.to raise_error(ArgumentError)
```
luke-hill: can we pass a partial / full message into this.

spec/section_spec.rb
```diff
@@ -49,11 +100,43 @@ class PageWithAnonymousSection < SitePrism::Page
...
+      expect(Section.default_search_arguments).to be_nil
+    end
+
+    it { expect(Section).to respond_to(:set_default_search_arguments) }
+
+    it 'should return default search arguments' do
+      expect(BaseSection.default_search_arguments).to eql([:css, '.default'])
+    end
+
+    it 'should return only this section default if they are set' do
```
luke-hill: I'm confused by this name and what this test is doing different. If it's just testing the fact css / xpath are different can we use shared contexts?

spec/section_spec.rb
```diff
@@ -49,11 +100,43 @@ class PageWithAnonymousSection < SitePrism::Page
...
+      expect(ChildSection.default_search_arguments).to eql([:xpath, '//html'])
+    end
+
+    it 'should parent section default arguments if defaults are not set' do
```
luke-hill: returns default search arguments of the parent section if defaults are not set

spec/sections_spec.rb
```diff
@@ -22,6 +28,24 @@ class Page < SitePrism::Page
     it { is_expected.to respond_to(:has_single_section?) }
   end
 
+  describe '.section with class and block' do
+    before do
+      expect(subject).to receive(:find_first).and_return(:element)
```
luke-hill: should this be allow?

spec/sections_spec.rb
```diff
@@ -22,6 +28,24 @@ class Page < SitePrism::Page
...
+    end
+
+    it 'should be an instance of provided section class' do
```
luke-hill: I'm confused by what this test is verifying?

README.md
```diff
@@ -804,12 +804,44 @@ SitePrism allows adding sections to sections) is to call the `section`
...
+end
+```
+
+The CSS selector can be omitted if you are re-using the same locator for all
+references to the section Class. In order to do this, simply tell SitePrism that
+you want to use a default search argument.
+
+```ruby
+class People < SitePrism::Section
+  set_default_serach_arguments '.people'
```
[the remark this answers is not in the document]
ineverov: Oops, sorry.

README.md
```diff
@@ -804,12 +804,44 @@ SitePrism allows adding sections to sections) is to call the `section`
...
+class HomePage < SitePrism::Page
+  # section people_with_block will have headline and
+  # headline_clone elements in it
+  section :people_with_block, People do
+    element :headline_clone, 'h2'
+  end
+end
+```
+
+The CSS selector can be omitted if you are re-using the same locator for all
```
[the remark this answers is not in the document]
ineverov: 👍

README.md
```diff
@@ -804,12 +804,44 @@ SitePrism allows adding sections to sections) is to call the `section`
 method. It takes 3 arguments: the first is the name of the section as
 referred to on the page (sections that appear on multiple pages can be
 named differently). The second argument is the class of which an
-instance will be created to represent the page section, and the third
-argument is a css selector that identifies the root node of the section
+instance will be created to represent the page section, and the following
+arguments are [Capybara::Node::Finders](https://www.rubydoc.info/github/teamcapybara/capybara/master/Capybara/Node/Finders)
+arguments that identifies the root node of the section
```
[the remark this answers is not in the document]
ineverov: 👍

spec/section_spec.rb
ineverov: Will do

spec/sections_spec.rb
ineverov: allow will not make assertion method is called

spec/sections_spec.rb
ineverov: So, it is creating an anonymous class [here](https://github.com/natritmeyer/site_prism/pull/250/files#diff-0142fe7beaa583d4ce8030ab2a152817R242) with parent SingleSection and I want to check it

luke-hill:
Merged. Thankyou for your contribution

PR #252, opened 2018-05-08
Page#elements_present

luke-hill:
This new class instance method is a sister method to `#all_there?`

This method will instead return you every objects that is mapped AND present on the page.

I've also slightly refactored the `ElementChecker` autoloaded Module, so that the code reads a little nicer and is more extensible.

EDIT: Ignore the location / structure / nuances of the spec file. This will be fixed up and migrated into a better construct once a few other tickets have hit the codebase

EDIT 2: Don't worry about spec failure. Just a rubocop failure. Again will be tidied up once other tickets have landed.


luke-hill:
cc/ @tgaff @TheMetalCode @RustyNail - Ready for review

lib/site_prism/element_checker.rb
```diff
@@ -3,15 +3,29 @@
 module SitePrism
   module ElementChecker
     def all_there?
-      elements_to_check.all? { |element| send "has_#{element}?" }
+      elements_to_check.all? { |element| present?(element) }
```
ineverov: some syntax sugar?
```ruby
elements_to_check.all?(&method(:present?))
```

spec/new_temp_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      expected_elements :bob
+    end
+
+    before do
+      allow(page).to receive(:present?).with(:bob).and_return(true)
+      allow(page).to receive(:present?).with(:dave).and_return(false)
+      allow(page).to receive(:present?).with(:success_msg).and_return(true)
+    end
+
+    let(:page) { PageWithAFewElements.new }
```
ineverov: I'd move `let` and `subject` before `before` block

lib/site_prism/element_checker.rb
```diff
@@ -3,15 +3,29 @@
...
+    end
+
+    def elements_present
+      mapped_items.select { |element| present?(element) }
```
ineverov:
```ruby
mapped_items.select(&method(:present?))
```

lib/site_prism/element_checker.rb
luke-hill: Genuine question. Which is more common / preferred? I never knew you could do this, I thought you had to write a proc.

You've probably got more experience than me, so not fussed either way. I personally haven't seen this used very often. So don't know.

lib/site_prism/element_checker.rb
```diff
@@ -3,15 +3,29 @@
...
+      else
+        mapped_items
+      end
+    end
+
+    def mapped_items
+      self.class.mapped_items.uniq
+    end
+
+    def present?(element)
```
Systho: `present?` is already a method added by activesupport on all object as the opposite of `blank?`  ( /lib/active_support/core_ext/object/blank.rb )

I do not think it is a good idea to shadow this public method by a private one.

luke-hill: Hi there. Valid point and I think it's worth a change as it's a simple bugfix ticket.

If you have the capacity, feel free to raise a PR. Otherwise I'll get round to it as and when.

Can I ask if you're using a late(r), version of site_prism and capybara. As Ideally I'd just want this hitting the 3.x series (As backporting then just means extra code / gem releases) - Are you happy for it just to hit the 3.x series?

Systho: I hit the problem on 2.17 but it's ok I have a small `site_prism_patch.rb` loaded by my capybara initializer :)

I'll try to make a PR against  master.

PS : I 'm taking the opportunity of this ticket to say a HUUUUGE thank you to you guys ! I've been using site_prism  since 1.2 and advocating everyone to use it. Adopting it is -  hands down - the best refactoring every single capybara test suite should make.

PR #254, opened 2018-05-09
Fix references to results_page in README

robd:


luke-hill:
Merged. Thanks for your contribution

PR #256, opened 2018-05-14
Rubocop fixes

ineverov:
Fixed line length offenses for `lib/` (except `element_container.rb` -  there are several PR changing this file)

### DIFF
--- a/lib/site_prism/addressable_url_matcher.rb
+++ b/lib/site_prism/addressable_url_matcher.rb
@@ -5,11 +5,10 @@ require 'base64'
 
 module SitePrism
   class AddressableUrlMatcher
-    COMPONENT_NAMES = %i[scheme user password host port path query fragment].freeze
-    COMPONENT_PREFIXES = {
-      query: '?',
-      fragment: '#'
-    }.freeze
+    COMPONENT_NAMES = %i[scheme user password host
+                         port path query fragment ].freeze
+
+    COMPONENT_PREFIXES = { query: '?', fragment: '#' }.freeze
 
     attr_reader :pattern
 
@@ -17,7 +16,8 @@ module SitePrism
       @pattern = pattern
     end
 
-    # @return the hash of extracted mappings from parsing the provided URL according to our pattern,
+    # @return the hash of extracted mappings from
+    # parsing the provided URL according to our pattern,
     # or nil if the URL doesn't conform to the matcher template.
     def mappings(url)
       uri = Addressable::URI.parse(url)
@@ -31,15 +31,15 @@ module SitePrism
       result
     end
 
-    # Determine whether URL matches our pattern, and optionally whether the extracted mappings match
-    # a hash of expected values.  You can specify values as strings, numbers or regular expressions.
+    # Determine whether URL matches our pattern, and
+    # optionally whether the extracted mappings match
+    # a hash of expected values.  You can specify values
+    # as strings, numbers or regular expressions.
     def matches?(url, expected_mappings = {})
       actual_mappings = mappings(url)
-      if actual_mappings
-        expected_mappings.empty? || all_expected_mappings_match?(expected_mappings, actual_mappings)
-      else
-        false
-      end
+      return false unless actual_mappings
+      expected_mappings.empty? ||
+        all_expected_mappings_match?(expected_mappings, actual_mappings)
     end
 
     private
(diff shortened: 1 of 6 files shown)


lib/site_prism/addressable_url_matcher.rb
```diff
@@ -72,29 +72,33 @@ def extract_component_templates
...
+    # Returns empty hash if the template omits the component,
+    # a set of substitutions if the
+    # provided URI component matches the template component,
+    # or nil if the match fails.
     def component_matches(component, uri)
-      extracted_mappings = {}
-      component_template = component_templates[component]
-      if component_template
+      mappings = {}
+      if (component_template = component_templates[component])
```
luke-hill: I personally dislike this kind of assignment, so I altered it to the previous one

lib/site_prism/addressable_url_matcher.rb
```diff
@@ -72,29 +72,33 @@ def extract_component_templates
...
         component_url = uri.public_send(component).to_s
-        extracted_mappings = component_template.extract(component_url)
-        unless extracted_mappings
+        mappings = component_template.extract(component_url)
+        unless mappings
           # to support Addressable's expansion of queries
           # ensure it's parsing the fragment as appropriate (e.g. {?params*})
           prefix = COMPONENT_PREFIXES[component]
-          return nil unless prefix && (extracted_mappings = component_template.extract(prefix + component_url))
+          return nil unless prefix
```
luke-hill: Is this line the same (And some weird ruby-ness I don't know)

lib/site_prism/addressable_url_matcher.rb
```diff
@@ -115,28 +119,32 @@ def substitutions
...
 
     def reverse_substitutions
-      @reverse_substitutions ||= slugs.each_with_index.reduce({}) do |memo, slug_index|
-        slug, index = slug_index
-        memo.merge(slug_prefix(slug) + substitution_value(index) => slug, substitution_value(index) => slug)
-      end
+      @reverse_substitutions ||=
+        slugs.each_with_index.reduce({}) do |memo, slug_index|
+          slug, index = slug_index
+          memo.merge(slug_prefix(slug) + substitution_value(index) => slug,
```
luke-hill: if we're expanding the brackets could we go down the
```
method(
  stuff
)
```

route

lib/site_prism/exceptions.rb
```diff
@@ -6,7 +6,8 @@ class NoUrlMatcherForPage < StandardError; end
 
   class InvalidUrlMatcher < StandardError
     def message
-      'Could not automatically match your URL. Templated port numbers are unsupported.'
+      "Could not automatically match your URL. \
+Templated port numbers are unsupported."
```
luke-hill: this needs indenting once from the previous line

lib/site_prism/loadable.rb
```diff
@@ -35,24 +39,30 @@ def self.included(base)
...
+      previously_loaded = loaded
+      message = 'A block was expected, but none received.'
+      raise ArgumentError, message unless block_given?
 
       # Within the block, cache loaded? to optimize performance.
       self.loaded = loaded?
       unless loaded
-        message = "Failed to load because: #{load_error || 'no reason given'}"
-        raise(::SitePrism::NotLoadedError, message)
+        raise(::SitePrism::NotLoadedError,
```
luke-hill: lets keep this one as per previous to keep the consistency

lib/site_prism/addressable_url_matcher.rb
ineverov: I was trying to leverage between line length and method length. Currently, we have 80 chars in 11 (code) lines

lib/site_prism/addressable_url_matcher.rb
ineverov: original condition `return nil if prefix is nil` or `component_template.extract(prefix + component_url) is nil`. In the first case we can immediately return nil (new line 93). in second case mapping will be nil and we return nil at the end of the method (line 97)

lib/site_prism/addressable_url_matcher.rb
ineverov: yes, sure we can, but again we have cop to check method length < 11 lines

lib/site_prism/exceptions.rb
ineverov: in this case, you'll have extra spaces in the message.
Now,` "Could not automatically match your URL. Templated port numbers are unsupported."`
if ident `"Could not automatically match your URL.          Templated port numbers are unsupported."`

lib/site_prism/loadable.rb
ineverov: again leverage between line and method length... :(

lib/site_prism/addressable_url_matcher.rb
luke-hill: Yes... we can... Top spot!

lib/site_prism/addressable_url_matcher.rb
luke-hill: Shouldn't be affected.

luke-hill:
As we've cut the 2.13 gem release can we get this one rebased and fixed up (As it's a simple fix and I just wanna get this in) @ineverov

ineverov:
@luke-hill rebased

lib/site_prism/addressable_url_matcher.rb
luke-hill: Can we revert this one?

lib/site_prism/addressable_url_matcher.rb
luke-hill: And alert this one as well

lib/site_prism/loadable.rb
luke-hill: Can we revert this one

lib/site_prism/page.rb
```diff
@@ -10,7 +10,8 @@ class Page
     extend ElementContainer
 
     load_validation do
-      [displayed?, "Expected #{current_url} to match #{url_matcher} but it did not."]
+      [displayed?,
```
luke-hill: for Arrays I think having
```
[
  val1,
  val2,
]
```

Looks the cleanest

ineverov:
@luke-hill Fixed comments. Going to work on element container soon

lib/site_prism/page.rb
```diff
@@ -10,7 +10,10 @@ class Page
     extend ElementContainer
 
     load_validation do
-      [displayed?, "Expected #{current_url} to match #{url_matcher} but it did not."]
+      [
```
ineverov: This change made Page class longer than 100 lines. This is the reason rubocop is failing

Issue #257, opened 2018-05-18
Use of Timeout.timeout is a problem


twalpole:
Because of the way it works `Timeout.timeout` is the most dangerous method in Ruby and should only be used around code that is trivial, completely understood and guaranteed to be safe for a random exception to occur at any point in its execution, or that blocks those exceptions using `Thread.handle_intterupt`.  SitePrism has two places where it wraps Capybara calls with `Timeout.timeout` including
https://github.com/natritmeyer/site_prism/blob/9c39f4b3b8ba47427a34b7b0cccd2a527c920f51/lib/site_prism/element_container.rb#L180 which can cause drivers to end up in unrecoverable states if the timeout ever occurs.

twalpole:
Note: this still has the issue that the `find_args` and `runtime_args` won't get combined correctly, but that's a problem in all the ElementContainer methods and should probably be fixed in a separate PR

luke-hill:
Closing in favour of PR

PR #258, opened 2018-05-18
Remove use of Timeout.timeout and fix merging of find_args and runtime_args - Fixes Issue #257

twalpole:
This removes the use of Timeout.timeout and also fixes the issue where the visibility/invisibility waiters would let through exceptions Capybara would normally retry within the time limit (StaleElementReferenceError, etc)


twalpole:
Note: this still has the issue that the find_args and runtime_args won't get combined correctly, but that's a problem in all the ElementContainer methods and should probably be fixed in a separate PR

twalpole:
The second commit here fixes the issue of combining the find_args and runtime_args with extra options when specified that would end up losing options in the find_args and/or runtime_args

luke-hill:
I have no issue with this and think it seems ok. But being brutally honest I'd need to look into this in detail as some of the code has gone a bit over me.

I'll definitely try put aside some time either tomorrow or on the weekend. Thanks for the help though, it's greatly appreciated.

twalpole:
@luke-hill No problem, if you want explanation for any specific part just ask.  A quick explanation for the find/runtime/extra args thing is that if you define an element with options

    element :info, :field, 'Information', type: 'textarea'

and then try to call any of the methods with runtime options

   wait_until_info_visible(with: 'My Info')

the current code would not combine the options correctly and end up at best just ignoring the initial element definitions `type` option, and in this case would ignore both the `type` option and the `with` runtime option because `wait_until_info_visible` passes its own option. So the options passed through actually become  `:field, 'Information, { type: 'texturea' }, {with: 'My Info'}, {visible: true}`  which obviously isn't correct since it should be `:field, 'information', {type: 'textarea', with: 'My Info', visible: true}` for Capybara to handle it correctly.

luke-hill:
If you're on either cucumber/selenium slack's I'm on there as "luke". If you've got any time this week to chat privately about this.

From what you've said it seems simple enough.

Ping @tgaff / @ineverov as you'll likely understand the ruby more than me.

.ruby-version
```diff
@@ -1 +0,0 @@
-ruby-2.2
```
luke-hill: can we leave this in for now. Not quite sure what to do going forward with this. I introduced it to try stabilise the old stuff, might get rid of it soon

features/step_definitions/elements_steps.rb
```diff
@@ -16,6 +16,11 @@
   expect(@test_site.home.people).to have_individuals(count: 4)
 end
 
+Then('I can see optioned individual people in the people list') do
+  expect(@test_site.home.people.optioned_individuals.size).to eq(4)
+  expect(@test_site.home.people).to have_optioned_individuals(count: 4)
```
luke-hill: can we have a linespace between expectations

lib/site_prism/element_container.rb
```diff
@@ -85,6 +85,16 @@ def raise_wait_for_no_if_failed(obj, name, timeout, failed)
             "Timed out after #{timeout}s waiting for no #{obj.class}##{name}"
     end
 
+    def merge_args(find_args, runtime_args, override_options = {})
+      find_args = find_args.dup
+      runtime_args = runtime_args.dup
```
luke-hill: When you duplicate these, won't they self-overwrite as they have the same name?

.ruby-version
twalpole: Ok -- but 2.2 is a weird choice since it is EOL

lib/site_prism/element_container.rb
twalpole: No - at first `find_args` refers to the passed in array.  If one modified that array (via pop) it would modify the array in the scope that called `merge_args`.  By calling `dup` on that array, `find_args` in the `merge_args` scope now refers to a new array and can modify it at will without affecting the array from the outer scope.

.ruby-version
luke-hill: Agreed. I am sorry. It is something reasonably high on my list. It's also one of the few things I can actually help with.

lib/site_prism/element_container.rb
luke-hill: This is really interesting. But I do feel it's complex. But thanks for the explanation. That does make sense!

So if you pass in [1,2,3,4] and then duplicate it. When you call [1,2,3,4].pop, the value of the array after this method has completed is still [1,2,3,4] just the array inside the method would be [1,2,3] right?

luke-hill:
Merged, thanks for your contribution

PR #261, opened 2018-05-25
Gemspec update

luke-hill:
Also enables Rubocop tests to pass up to V55


luke-hill:
@RustyNail This allows up to version 55 of rubocop to be used (And be fully rubocop compliant)

If you're happy with this I'll merge it in.

RustyNail:
@luke-hill 
I'm sorry for replying late.
I confirmed that it works well with the latest Rubocop (v 0.56.0) even here.
Looks good to me!

PR #262, opened 2018-05-25
TODO Cleanup

luke-hill:
Spring clean the first half of the page_spec file (Just mini-bit of DRYing / syntax, not placement)
Fix up weird code-coverage issue that was bugging me.


luke-hill:
Ping @tgaff @TheMetalCode if you've got time.

Issue #269, opened 2018-06-12
Wait for/until methods do not actually loop. Just perform check one time. 


zwarburg:
## Environment
- **Operating System**: Windows 10
- **What Gems / Versions you are using**: Capybara 2.18.0, SitePrism 2.13 (tried both the release and current master)
- **How you're interacting with the Framework**: Rspec Tests
- Browser/s Driver/s being used: Chrome with latest version and selenium-webdriver 3.9.0

## Expected Behavior
Waiting for element should continue as soon as the element appears on the screen. 

## Actual Behavior
Waits until the max timeout has been reached then moves forward regardless of whether the condition has been satified. 

When looking at the code for the waiter there is no loop, it only performs the check one time and then just waits until the timeout is reached. Waiter.wait_until_true is defined but doesn't appear to ever be used in the waiters. 

https://github.com/natritmeyer/site_prism/blob/master/lib/site_prism/element_container.rb#L158

luke-hill:
After an initial glance we seem to have almost every spec / feature covering this area around your specific use case. But "not" one specific to the timing you're mentioning. Waiting the full duration on a positive wait.

I'll investigate a bit further. But I'm reasonably confident I've seen this working well (Albeit I only use small timeouts).

Can you create a SSCCE for me to investigate the current codebase waiting too long. Or alternatively you can add a PR with a single commit adding a new spec that fails.

luke-hill:
After investigation. I can conclude that this issue doesn't occur in the master branch nor does it occur in the released gem. If you find information to the contrary can you open a new issue with a reproducible test case.

The code you refer to is calling a raw Capybara method. Which in turn will implicitly wait and auto-repoll for the condition to be passed. We explicitly pass in `0` in instances where we don't want this value to be used (See usage of `SitePrism.use_implicit_waits` configuration here: https://github.com/natritmeyer/site_prism/blob/master/README.md#using-capybara-implicit-waits)

In deference to the issues you mentioned not being explicitly integration tested. I have added https://github.com/natritmeyer/site_prism/commit/2fb4e45824971984f459837c8bd3ca420771a5db to cover these edge cases and prove that the timings (Both positive and negative), fall within accepted tolerances (For positive waiting it will continue once satisfied after passing. For negative tests it will continue around 0.1s after the completion of the timeout.

PR #270, opened 2018-06-13
Fixing waiters to use 'wait_until_true' method

zwarburg:
My attempt at fixing issue https://github.com/natritmeyer/site_prism/issues/269. This makes use of the `wait_until_true` method so that actual looping occurs.


luke-hill:
Will Explain changes made to codebase in separate issue.

PR #273, opened 2018-06-20
Rubocop. Part 3

ineverov:


lib/site_prism/element_container.rb
```diff
@@ -68,16 +73,17 @@ def section(section_name, *args, &block)
         section_class, find_args = extract_section_options(args, &block)
         build(section_name, *find_args) do
           define_method section_name do |*runtime_args, &runtime_block|
-            section_class.new self, find_first(*merge_args(find_args, runtime_args)), &runtime_block
+            section_element = find_first(*merge_args(find_args, runtime_args))
+            section_class.new self, section_element, &runtime_block
```
luke-hill: can we wrap the creation in `()`

lib/site_prism/element_container.rb
```diff
@@ -68,16 +73,17 @@ def section(section_name, *args, &block)
...
         end
       end
 
-      def sections(section_collection_name, *args, &block)
+      def sections(collection_name, *args, &block)
         section_class, find_args = extract_section_options(args, &block)
-        build(section_collection_name, *find_args) do
-          define_method(section_collection_name) do |*runtime_args, &element_block|
-            raise_if_block(self, section_collection_name.to_s, !element_block.nil?)
+        build(collection_name, *find_args) do
```
luke-hill: should we call this either `section_name` or even better just `name` because we know it's a section

lib/site_prism/element_container.rb
```diff
@@ -154,7 +160,7 @@ def create_nonexistence_checker(element_name, *find_args)
         method_name = "has_no_#{element_name}?"
         create_helper_method(method_name, *find_args) do
           define_method(method_name) do |*runtime_args|
-            wait_time = SitePrism.use_implicit_waits ? Capybara.default_max_wait_time : 0
+            wait_time = SitePrism.use_implicit_waits ? max_wait_time : 0
```
luke-hill: How about we extract this line out to new Line 11. And switch it from a ternary to if/else?

lib/site_prism/page.rb
```diff
@@ -72,16 +72,18 @@ def template_backed_matches
       matcher_template.mappings(page.current_url)
     end
 
+    # rubocop:disable Naming/AccessorMethodName
```
luke-hill: lets just leave this as a regular rubocop override. As we might provide other `setters` later on in the projects life.

lib/site_prism/section.rb
```diff
@@ -52,7 +54,9 @@ def evaluate_script(input)
 
     def parent_page
       candidate_page = parent
-      candidate_page = candidate_page.parent until candidate_page.is_a?(SitePrism::Page)
+      begin
```
luke-hill: Why don't we just go back to what was used previously which was `until` at the top. I think it reads nicer.

spec/loadable_spec.rb
```diff
@@ -12,7 +12,8 @@
 
   describe 'class methods' do
     describe '#load_validations' do
-      it 'returns the load_validations from the current and all ancestral classes in a hierarchical, defined order' do
+      it "returns the load_validations from the current \
```
luke-hill: can we just slightly modify this to say `returns the load_validations from the current class`

spec/loadable_spec.rb
```diff
@@ -62,12 +68,16 @@
       expect(james_bond).not_to have_received(:drink_martini)
     end
 
-    it 'raises an exception with a specific error message if available when a load validation fails' do
+    it "raises an exception with a specific error message \
+if available when a load validation fails" do
```
luke-hill: remove the words `if available`

spec/loadable_spec.rb
```diff
@@ -62,12 +68,16 @@
...
       loadable.load_validation { [false, 'all your base are belong to us'] }
 
       expect do
         loadable.new.when_loaded { :foo }
-      end.to raise_error(SitePrism::NotLoadedError, /all your base are belong to us/)
+      end.to raise_error(
+        SitePrism::NotLoadedError,
+        /all your base are belong to us/
```
luke-hill: this can be altered to be an exact match instead of regex match

spec/loadable_spec.rb
```diff
@@ -150,7 +160,8 @@
       expect(inheriting_loadable.new).not_to be_loaded
     end
 
-    it 'returns false if any load validation fails at any point in the inheritance chain' do
+    it "returns false if any load validation fails at any point \
```
luke-hill: just change this entire message to `it 'returns false is any load validation fails'` Don't need the rest of the message. Plus single quotes.

.rubocop.yml
```diff
@@ -4,22 +4,7 @@ AllCops:
 Documentation:
   Enabled: false
 
-AccessorMethodName:
```
luke-hill: These probably need re-adding in for now.

lib/site_prism/section.rb
```diff
@@ -11,9 +11,11 @@ class Section
 
     attr_reader :root_element, :parent
 
+    # rubocop:disable Naming/AccessorMethodName
```
luke-hill: Lets remove this override and leave the rubocop rule as ignored in the suite (We may start introducing more)

lib/site_prism/section.rb
luke-hill: Can we revert this one for now please

spec/loadable_spec.rb
luke-hill: Can we fix this item up whilst we're here

spec/loadable_spec.rb
```diff
@@ -150,7 +158,7 @@
       expect(inheriting_loadable.new).not_to be_loaded
     end
 
-    it 'returns false if any load validation fails at any point in the inheritance chain' do
+    it 'returns false is any load validation fails' do
```
luke-hill: if*

spec/page_spec.rb
```diff
@@ -122,7 +126,9 @@ def foo?
 
       context 'With Passing Load Validations' do
         it 'executes the block' do
-          expect(page_with_load_validations.load { :return_this }).to eq(:return_this)
+          expect(
```
luke-hill: lets put this onto 2 lines as per other ones

spec/page_spec.rb
```diff
@@ -134,7 +140,9 @@ def foo?
 
       context 'With Failing Load Validations' do
         it 'raises an error' do
-          allow(page_with_load_validations).to receive(:must_be_true).and_return(false)
+          allow(
```
luke-hill: and this one onto 2 lines

spec/page_spec.rb
```diff
@@ -157,9 +165,12 @@ class PageToSetUrlMatcherAgainst < SitePrism::Page
...
   end
 
-  it 'should raise an exception if displayed? is called before the matcher has been set' do
+  it "should raise an exception if displayed? \
+is called before the matcher has been set" do
     class PageWithNoMatcher < SitePrism::Page; end
-    expect { PageWithNoMatcher.new.displayed? }.to raise_error SitePrism::NoUrlMatcherForPage
+    expect do
+      PageWithNoMatcher.new.displayed?
+    end.to raise_error SitePrism::NoUrlMatcherForPage
```
luke-hill: can we wrap the error in `()` after raise_error

spec/page_spec.rb
```diff
@@ -180,31 +191,42 @@ class PageWithBogusFullUrlMatcher < SitePrism::Page
...
+          "Could not automatically match your URL. \
+Templated port numbers are unsupported."
+        )
     end
   end
 
   describe 'with a full string URL matcher' do
     class PageWithStringFullUrlMatcher < SitePrism::Page
-      set_url_matcher 'https://joe:bump@bla.org:443/foo?bar=baz&bar=boof#myfragment'
+      set_url_matcher \
```
luke-hill: can we revert this one for now. I don't like how we're already stretching the line before defining anything

spec/section_spec.rb
```diff
@@ -31,11 +31,15 @@ class PageWithSections < SitePrism::Page
       end
 
       it 'should be an instance of provided section class' do
-        expect(subject.section_with_a_block.class.ancestors).to include(SingleSection)
+        expect(
```
luke-hill: can we put this on 2lines

spec/section_spec.rb
```diff
@@ -31,11 +31,15 @@ class PageWithSections < SitePrism::Page
...
+          subject.section_with_a_block.class.ancestors
+        ).to include(SingleSection)
       end
 
       it 'should have elements from the base section' do
-        expect(subject.section_with_a_block).to respond_to(:single_section_element)
+        expect(
```
luke-hill: This one also

spec/section_spec.rb
```diff
@@ -133,7 +143,8 @@ class SectionWithDefaultArgumentsForParent < SectionWithDefaultArguments; end
         end
       end
 
-      context 'with niether default search arguments nor search arguments provided' do
+      context "with niether default search arguments \
```
luke-hill: * `neither`

typo that has been left in for a while evidently

spec/section_spec.rb
```diff
@@ -176,12 +189,16 @@ class SecondChildSection < BaseSection; end
...
+    it "should return only this section's \
+default search arguments if they are set" do
       expect(ChildSection.default_search_arguments).to eql([:xpath, '//html'])
     end
 
-    it "should return parent section's default search arguments if defaults are not set" do
-      expect(SecondChildSection.default_search_arguments).to eql([:css, '.default'])
+    it "should return parent section's default \
+search arguments if defaults are not set" do
+      expect(
```
luke-hill: this can live on 2 lines

spec/section_spec.rb
```diff
@@ -246,7 +263,9 @@ class SecondChildSection < BaseSection; end
 
   describe '#evaluate_script' do
     it 'delegates through Capybara' do
-      expect(Capybara.current_session).to receive(:evaluate_script).with('How High?')
+      expect(
```
luke-hill: This can also live on 2 lines

spec/section_spec.rb
```diff
@@ -296,7 +315,12 @@ class SecondChildSection < BaseSection; end
     context 'when root element is nil' do
       let(:root_element) { nil }
 
-      before { allow(Capybara).to receive(:current_session).and_return('current session') }
+      before do
+        allow(Capybara).to(
```
luke-hill:
```
allow(Capybara)
  .to receive(:current_session)
  .and_return('current session')
```

spec/sections_spec.rb
```diff
@@ -38,8 +39,12 @@ class Page < SitePrism::Page
     end
 
     it 'should use default arguments' do
-      expect(SitePrism::Section).to receive(:new).with(subject, :element1).ordered
-      expect(SitePrism::Section).to receive(:new).with(subject, :element2).ordered
+      expect(
```
luke-hill: Both of these can live on 2 lines

spec/page_spec.rb
luke-hill: Can this go onto 2 lines

spec/page_spec.rb
luke-hill: Can this go onto 2 lines

spec/page_spec.rb
luke-hill: :point_up:

spec/page_spec.rb
luke-hill: :point_up:

spec/page_spec.rb
```diff
@@ -339,7 +377,7 @@ class PageWithParameterizedUrlMatcher < SitePrism::Page
 
   describe 'with a regexp matcher' do
     class PageWithRegexpUrlMatcher < SitePrism::Page
-      set_url_matcher(/foos\/(\d+)/)
+      set_url_matcher(%r{foos\/(\d+)})
```
luke-hill: I actually disagree with this cop and ban it from other suites. Can we revert this and just have it as an override

spec/section_spec.rb
luke-hill: :point_up:

spec/section_spec.rb
luke-hill: :point_up:

lib/site_prism/element_container.rb
```diff
@@ -178,41 +184,44 @@ def create_waiter(element_name, *find_args)
...
         method_name = "wait_until_#{element_name}_visible"
         create_helper_method(method_name, *find_args) do
-          define_method(method_name) do |timeout = Capybara.default_max_wait_time, *runtime_args|
-            unless element_exists?(*merge_args(find_args, runtime_args, visible: true, wait: timeout))
-              raise SitePrism::TimeOutWaitingForElementVisibility
-            end
+          define_method(method_name) do |timeout = max_wait_time, *runtime_args|
+            visibility_args = { visible: true, wait: timeout }
+            args = merge_args(find_args, runtime_args, **visibility_args)
+            next if element_exists?(*args)
```
luke-hill: should this be `return` - I know how `next` works but we're not quite in an enumerable.? ignore if not relevant

Same for item below on line215

ineverov: nice catch, probably `return` will be better here since this is a method definition.
I was confused because `return` is used only in methods. But we are defining method here.

PR #277, opened 2018-06-28
Capybara 3 rework

### LINKED ISSUE #243: element(s) use implicit waits regardless of setting
## Summary
`@search_page.search_results` waits for Capybara.default_max_wait_time when `SitePrism.use_implicit_waits = false`

## Environment
- Operating System (Mac)
- site_prism master at c9526bf9fe38039973ed339d6fbb05a03deb7b20     capybara (2.18.0)
- How you're interacting with the Framework - Using the builtin cucumber tests
- Firefox 59.0.2

## Expected Behavior
Implicit waits should not be performed when disabled by `config.use_implicit_waits = false`

### Detail
Per the readme:
> By default, SitePrism element and section methods do not utilize
> Capybara's implicit wait methodology and will return immediately if
> the element or section requested is not found on the page.  Add the
> following code to your spec_helper file to enable Capybara's implicit
> wait methodology to pass through:
> 
> ```ruby
> SitePrism.configure do |config|
>   config.use_implicit_waits = true
> end
> ```
> 
> This enables you to replace this:
> 
> ```ruby
> # wait_until methods always wait for the element to be present on the page:
> @search_page.wait_for_search_results
> 
> # Element and section methods do not:
> @search_page.search_results
> ```
> 

with this:

```ruby
# With implicit waits enabled, use of wait_until methods is no longer required. This method will
# wait for the element to be found on the page until the Capybara default timeout is reached.
@search_page.search_results
```

## Actual Behavior
For something like `@search_page.search_results` implicit waits **are** performed.

## Steps to reproduce

Add the following cucumber steps:

```rb
+
+Then(/^implicit waits are not enabled$/) do
+  expect(SitePrism.use_implicit_waits).to eq false
+end
+
+Then(/^the slow element is not waited for$/) do
+  start_time = Time.now
+  @test_site.home.some_slow_element
+  expect(Time.now - start_time).to be < 0.5
+end


# and features/waiting.feature
+  Scenario: Elements Are Not Waited For By
+    Then the slow element is not waited for
+    And implicit waits are not enabled

```

Run `bundle exec cucumber`

Observe: 

```
  Scenario: Elements Are Not Waited For By  # features/waiting.feature:53
    When I navigate to the home page        # features/step_definitions/navigation_steps.rb:3
    Then the slow element is not waited for # features/step_definitions/page_element_interaction_steps.rb:169
      expected: < 0.5
           got:   2.034348 (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/page_element_interaction_steps.rb:172:in `/^the slow element is not waited for$/'
      features/waiting.feature:55:in `Then the slow element is not waited for'
    And implicit waits are not enabled      # features/step_definitions/page_element_interaction_steps.rb:164
```

See also: https://github.com/natritmeyer/site_prism/commit/7c59938cc91bb6ad0b22c18861227f0dc2479d67

The above indicates that despite implicit waits being **disabled** they were in fact used.  (some_slow_element waits 2s before appearing on page.)

luke-hill:
Enable provisional Capybara 3 support.
- Fix up method collisions and amend specs as appropriate
- Fix newly introduced `#all` functionality (We don't call `#first`)
- Deprecate `collection` DSL definition

EDIT: Additional work that is also covered here.
- Reduced Travis test overhead by 20% (Removing Ruby 2.1 support, which in theory is impossible with Capybara3 and we shouldn't test as it's our lowest level)
- Fix up of codecoverage back up to ~100% (There are 3 lines uncalled which are the `collection` deprecation - This proves it was never actually called in our codebase internally)
- Fix up implicit waiting for all methods. This closes #243

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,7 @@
 AllCops:
   TargetRubyVersion: 2.2
+  Exclude:
+    - './spec/*_spec.rb'
 
 Documentation:
   Enabled: false
@@ -13,13 +15,9 @@ Style/SafeNavigation:
 Style/RegexpLiteral:
   Enabled: false
 
-BlockLength:
-  Exclude:
-    - './spec/*_spec.rb'
-
-# Offense count: 24
+## Offense count: 13
 Metrics/LineLength:
-  Max: 107
+  Max: 99
 
 Gemspec/RequiredRubyVersion:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,7 +17,7 @@ BlockLength:
   Exclude:
     - './spec/*_spec.rb'
 
-# Offense count: 18
+# Offense count: 24
 Metrics/LineLength:
   Max: 107
 
--- a/lib/site_prism/element_container.rb
+++ b/lib/site_prism/element_container.rb
@@ -46,33 +46,38 @@ module SitePrism
     module ClassMethods
       attr_reader :mapped_items, :expected_items
 
-      def element(element_name, *find_args)
-        build(element_name, *find_args) do
-          define_method(element_name.to_s) do |*runtime_args, &element_block|
-            raise_if_block(self, element_name.to_s, !element_block.nil?)
+      def element(name, *find_args)
+        build(name, *find_args) do
+          define_method(name) do |*runtime_args, &element_block|
+            raise_if_block(self, name, !element_block.nil?)
             _find(*merge_args(find_args, runtime_args))
           end
         end
       end
 
-      def elements(collection_name, *find_args)
-        build(collection_name, *find_args) do
-          define_method(collection_name.to_s) do |*runtime_args, &element_block|
-            raise_if_block(self, collection_name.to_s, !element_block.nil?)
+      def elements(name, *find_args)
+        build(name, *find_args) do
+          define_method(name) do |*runtime_args, &element_block|
+            raise_if_block(self, name, !element_block.nil?)
             _all(*merge_args(find_args, runtime_args))
           end
         end
       end
-      alias collection elements
+
+      def collection(name, *find_args)
+        warn 'Using collection is now deprecated and will be removed.'
+        warn 'Use elements DSL notation instead.'
+        elements(name, *find_args)
+      end
 
       def expected_elements(*elements)
         @expected_items = elements
       end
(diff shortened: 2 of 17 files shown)


luke-hill:
Can I get a review from @ineverov as well as @twalpole.
I've modified the `merge_args` method I "believe" correctly Thomas. But It's highly likely I've misunderstood some components.

@tgaff If you have capacity to review as well that would be awesome.

lib/site_prism/section.rb
```diff
@@ -62,16 +62,20 @@ def native
...
+      page.find(*find_args)
+    end
+
+    def _all(*find_args)
+      page.all(*find_args)
     end
 
     def element_exists?(*find_args)
-      root_element.has_selector?(*find_args) unless root_element.nil?
+      page && page.has_selector?(*find_args)
```
ineverov: Is it possible to have a section without root element?

lib/site_prism/element_container.rb
```diff
@@ -32,48 +32,66 @@ def raise_wait_for_no_if_failed(obj, name, timeout, failed)
...
     def merge_args(find_args, runtime_args, override_options = {})
       find_args = find_args.dup
       runtime_args = runtime_args.dup
       options = {}
       options.merge!(find_args.pop) if find_args.last.is_a? Hash
       options.merge!(runtime_args.pop) if runtime_args.last.is_a? Hash
       options.merge!(override_options)
+      options[:wait] = false unless wait_required?(options)
+
+      return [*find_args, *runtime_args] if options.empty?
```
ineverov: I guess it's not necessary

.rubocop.yml
```diff
@@ -18,7 +18,7 @@ BlockLength:
     - './spec/*_spec.rb'
     - 'site_prism.gemspec'
 
-# Offense count: 18
+# Offense count: 24
```
ineverov: Can you fix those new offenses if it is possible?

lib/site_prism/section.rb
luke-hill: These were just direct replacements. But no you cannot have a section without a root_element. I'll alter this.

luke-hill:
I've merged this in as is @twalpole. If I've made any glaring errors I'll try DM you and put them into another PR

Issue #227, opened 2018-03-19
Error when iframe is called without a block


mdesantis:
Hello! Thank you for this wonderful library!

I have a page with a payments form, with a radio button. When the user clicks the radio button, some iframes with the credit card details forms gets added to the page. I should fill the credit card number field within the iframe that gets loaded. This is the test code:

```ruby
page = Checkout::PaymentPage.new
page.payment_form.credit_card_payment_method_radio_button.click # loads the iframe
page.wait_for_credit_card_number_iframe
expect(page).to have_credit_card_number_iframe # passes
page.credit_card_number_iframe #raises an error
# Here I should call `page.credit_card_number_iframe` in order to set the credit card number
```

Which is causing the following error:

```
NoMethodError:
  undefined method `call' for nil:NilClass
  Did you mean?  caller
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/site_prism-2.11/lib/site_prism/element_container.rb:60:in `block (2 levels) in iframe'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/capybara-2.18.0/lib/capybara/session.rb:426:in `within_frame'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/capybara-2.18.0/lib/capybara/dsl.rb:50:in `block (2 levels) in <module:DSL>'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/site_prism-2.11/lib/site_prism/element_container.rb:59:in `block in iframe'
# ./spec/features/purchase_spec.rb:113:in `block (2 levels) in <top (required)>'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/rspec-core-3.7.1/lib/rspec/core/example.rb:254:in `instance_exec'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/rspec-core-3.7.1/lib/rspec/core/example.rb:254:in `block in run'
# /home/user/.asdf/installs/ruby/2.3.6/lib/ruby/gems/2.3.0/gems/rspec-core-3.7.1/lib/rspec/core/example.rb:500:in `block in with_around_and_singleton_context_hooks'
[...]

PR #281, opened 2018-07-06
Fix issue where iFrame is called without a block

luke-hill:
Fixes issue #227 raised by @mdesantis.

When an iFrame is defined as a Page Object it must be called with a block. Corresponding code and tests added to check for this.


luke-hill:
Ping @TheMetalCode @RustyNail can I get a quick review on this if you have time.

mdesantis:
For me it's great, if anyone cares :)

PR #285, opened 2018-07-12
Remove procedural block in page.rb

luke-hill:
Clears up one of the 3.0 roadmap comments.

Note that this WILL change the load order and put this item last, however as it's being deprecated / removed I don't foresee an issue with it.

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -10,11 +10,11 @@ module SitePrism
     include Loadable
     include ElementContainer
 
-    load_validation do
-      [
-        displayed?,
-        "Expected #{current_url} to match #{url_matcher} but it did not."
-      ]
+    # When instantiating the page. A single default validation will be added
+    # When calling #load, all of the validations set will be ran
+    # in order, with the default "displayed?" validation being ran first
+    def initialize
+      add_displayed_validation
     end
 
     def page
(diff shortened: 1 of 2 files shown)


luke-hill:
ping @ineverov @RustyNail for reviews

lib/site_prism/page.rb
```diff
@@ -10,11 +10,12 @@ class Page
...
-      [
-        displayed?,
-        "Expected #{current_url} to match #{url_matcher} but it did not."
-      ]
+    # When instantiating the page. A default validation will be added to all
+    # validations you define and add to the Page Class.
+    # When calling #load, all of the validations that are set will be ran
+    # in order, with the default "displayed?" validation being ran _LAST_
+    def initialize
+      add_displayed_validation
```
ineverov: It will add `displayed?` load validation to the class object each time you initialize the page.
I'd do the following
```
def self.add_displayed_validation
...
end
add_displayed_validation
```

luke-hill: Cheers. Let me look into this, I had a bit of a brainwave to add it elsewhere in the code, as I wanted to remove any procedural code. I also probably should add a unit test that re-initializing the class doesn't re-add the load proc.

Good spot. This needs fixing!

PR #289, opened 2018-07-13
Improve runtimes of cucumber tests by around 30%

luke-hill:
Have reduced JS delay times (90% of the improvements)
Reduced implicit wait timeouts (10%)
Only looked into one of the big pages. Also tidied up a little bit of naming conventions.


luke-hill:
ping @tgaff @TheMetalCode if you've got time to review can you take a look?

PR #295, opened 2018-07-23
Wait key refactor

luke-hill:
All wait keys are now assigned in the same way (Added new private methods to determine which wait time to use)
Fixed up a load of old language that described poorly named features/scenarios
Rubocop minor tidy (Only 7 linelength failures)

### DIFF
--- a/features/step_definitions/wait_steps.rb
+++ b/features/step_definitions/wait_steps.rb
@@ -71,7 +71,7 @@ Then("the element I am waiting for doesn't disappear in time") do
 end
 
 When('I wait for the section element that takes a while to appear') do
-  @test_site.section_experiments.parent_section.wait_for_slow_section_element
+  @test_site.section_experiments.parent_section.wait_for_slow_element
 end
 
 When('I wait for the section element that takes a while to disappear') do
@@ -82,7 +82,7 @@ end
 
 Then('the slow section appears') do
   expect(@test_site.section_experiments.parent_section)
-    .to have_slow_section_element
+    .to have_slow_element
 end
 
 Then('the removing section disappears') do
@@ -93,9 +93,9 @@ end
 Then("an exception is raised when I wait for a section that won't appear") do
   section = @test_site.section_experiments.parent_section
 
-  expect { section.wait_for_slow_section_element(0.25) }
+  expect { section.wait_for_slow_element(0.25) }
     .to raise_error(SitePrism::TimeOutWaitingForExistenceError)
-    .with_message('Timed out after 0.25s waiting for Parent#slow_section_element')
+    .with_message('Timed out after 0.25s waiting for Parent#slow_element')
 end
 
 Then("an exception is raised when I wait for a section that won't disappear") do
@@ -116,11 +116,11 @@ Then('the removing collection of sections disappears') do
 end
 
 When('I wait a variable time for elements to appear') do
-  @test_site.home.wait_for_lots_of_links(2.1)
+  @test_site.home.wait_for_lots_of_links(1.6)
 end
 
 When('I wait a variable time for elements to disappear') do
-  @test_site.home.wait_for_no_removing_links(2.1)
+  @test_site.home.wait_for_no_removing_links(1.6)
 end
 
 Then('I get a timeout error when I wait for an element that never appears') do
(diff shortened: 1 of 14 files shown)


luke-hill:
ping @ineverov - Fixed up the majority of rubocop offences, and made a few refactor changes that ticks off some TODO items. Next stop is working on the key-configuration of the default `loaded?` validation

lib/site_prism/element_container.rb
```diff
@@ -169,9 +177,9 @@ def create_existence_checker(element_name, *find_args)
         method_name = "has_#{element_name}?"
         create_helper_method(method_name, *find_args) do
           define_method(method_name) do |*runtime_args|
-            wait_time = SitePrism.use_implicit_waits ? max_wait_time : 0
-            visibility_args = { wait: wait_time }
-            element_exists?(*merge_args(find_args, runtime_args, **visibility_args))
+            visibility_args = { wait: checker_wait_time }
+            args = merge_args(find_args, runtime_args, **visibility_args)
```
tgaff: Curious, what's the double splat in all of these here for?

test_site/pages/section_experiments.rb
```diff
@@ -33,7 +33,8 @@ def downcase_title_text
 
   def cell_value=(value)
     execute_script(
-      "document.getElementById('first_search_result').children[0].innerHTML = '#{value}'"
+      "document.getElementById('first_search_result').children[0].innerHTML =
```
tgaff: Github uses a ~120 character limit on a 15" MBP; setting your limit to 120 would probably resolve these issues and potentially keep things more easily readable.

features/step_definitions/wait_steps.rb
```diff
@@ -116,13 +116,11 @@
 end
 
 When('I wait a variable time for elements to appear') do
-  @test_site.home.wait_for_lots_of_links
-  @test_site.home.wait_for_lots_of_links(0.1)
+  @test_site.home.wait_for_lots_of_links(2.1)
```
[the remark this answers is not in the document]
luke-hill: On the previous code it was waiting 2.1 seconds, so this is a like for like replacement. I've not looked into this, but obviously this will terminate earlier (As this waiter will pass in ~1s)

lib/site_prism/element_container.rb
```diff
@@ -8,10 +8,18 @@ def self.included(klass)
...
-    def max_wait_time
+    def waiter_wait_time
       Capybara.default_max_wait_time
     end
 
+    def checker_wait_time
+      if SitePrism.use_implicit_waits
+        waiter_wait_time
+      else
+        0
```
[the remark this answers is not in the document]
luke-hill: if you use a wait of 0, it doesn't wait, so it has the same logic. This is copied from the old line 172 ternary

lib/site_prism/element_container.rb
luke-hill: This is to retain the exploded hash

i.e. (Ripped from SO)

```
def foo(a, *b, **c)
  [a, b, c]
end


> foo 10
=> [10, [], {}]
> foo 10, 20, 30
=> [10, [20, 30], {}]
> foo 10, 20, 30, d: 40, e: 50
=> [10, [20, 30], {:d=>40, :e=>50}]
> foo 10, d: 40, e: 50
=> [10, [], {:d=>40, :e=>50}]
```

Essentially we've had a huge long standing bug, where multiple keyword arguments were being ignored (We only ever tested with 0 or 1), Thomas from Capybara fixed this up in the merge_args method

test_site/pages/section_experiments.rb
luke-hill: I have a "rough" plan of revisiting and tidying up all of the sample POM and JS / sample pages. One thing we can do is remove some duplication in the naming of things so we don't hit an issue where we get these stupidly long lines!

lib/site_prism/element_container.rb
tgaff: This is a little different though, isn't it?  This isn't using keyword args at all and `merge_args`'s declaration isn't the one doing the exploding.  The double-splat is in the method call, not declaration.  

```rb
def bar_args(find_args, runtime_args, override_options = {})
  [find_args, runtime_args, override_options]
end

# no splatter
> bar_args(10, 20, {a:1, b:2})
=> [10, 20, {:a=>1, :b=>2}]
# splatter
> bar_args(10, 20, **{a:1, b:2})
=> [10, 20, {:a=>1, :b=>2}]
> bar_args(10, 20, {a:1, b:2}) == bar_args(10, 20, **{a:1, b:2})
=> true
```

luke-hill: I'll look into this. I know this "works" but I'll remove that and see what happens. It might have been something I put in for `spec` passing. But not 100%

Give me a few days to triage.

luke-hill:
ping @tgaff Can you take another look at this. I didn't need the splatters there (So was just a mistake), and I've also improved some of the other items from CR (And a couple of linelength cops).

You ok for this to be merged?

tgaff:
Sorry @luke-hill I'm a little slow.  I'd already approved it though, I'm sure its fine.

PR #310, opened 2018-08-19
3.0 alpha release

luke-hill:
This is the parent branch all others are being raised against. So this will give us a CI indication of where we are up to (As the child branches don't get ran against Travis)

There isn't an indicative plan of when to cut this. I'll probably just cut it in September some time when I've done a reasonable amount of coding against it.


luke-hill:
Note to @ineverov / @tgaff - The `master` branch here now is being shaped as per v3 alpha work. As I've just cut a stable 2.16 release.

There are no immediate plans for any more feature / bug fix releases in the 2.x series. But if they're required I'll cut them. The plan is in the next 2-3 weeks to have a reasonably chunk (40%+) of the work done for v3 and I'll release an alpha version of the gem.

PR #312, opened 2018-08-22
Initial implicit waiting work

luke-hill:
This begins the work to migrate over implicit waits to being enabled by default. To start with I've just looked at doing a couple of things

- https://github.com/natritmeyer/site_prism/commit/66861dfecff23995ded8db527ec7ab76b29627d5 -> Remove the 2 waiter meta-programmed methods as these are no longer required. And fix up all `spec` files
- https://github.com/natritmeyer/site_prism/commit/2fd49314213162502d87be129189f9cadc48b027 -> Migrate over all `features` that used the old meta-programmed methods, either to use a visibility waiter or using the `wait` key assignment. I've added a separate ticket to re-visit these and ensure we're testing the key assignment in all methods. There are also other breakout tickets for reviewing the features to ensure they're all in the right place, and another to standardise the API


luke-hill:
ping @tgaff @ineverov - This "should" be ready to review now. I believe I've fixed everything that should make it work out of the box and configure implicit waits by default. But I've possibly missed something.

Also shrunk the helper DSL by 2 (wait_for_*), as these will now be used with regular locators (And failure to locate will revert to the Capybara DSL error codes, which we test for)

lib/site_prism/element_container.rb
```diff
@@ -185,32 +181,6 @@ def create_nonexistence_checker(element_name, *find_args)
         end
       end
 
-      def create_waiter(element_name, *find_args)
```
ineverov: Should we deprecate it first?

luke-hill: I’m in a couple of minds. But I’d prefer not to. 

What we can do though is deprecate it in the 2.x version. 

Reason why is I only want to do one set of big changes for v3. Not doing another major bump for a while. 

Thoughts @tgaff ??

In theory this release won’t be ready for a while so cutting another release that deprecates this seems valid?

luke-hill:
For the travis re-work @mnohai-mdsol feel free to raise a PR and do it there. As it's probably not relevant to this work.

features/step_definitions/wait_steps.rb
```diff
@@ -23,29 +23,26 @@
 Then("an exception is raised when I wait for an element that won't appear") do
   start_time = Time.now
 
-  expect { @test_site.home.wait_for_some_slow_element(1) }
-    .to raise_error(SitePrism::ExistenceTimeoutError)
-    .with_message('Timed out after 1s waiting for Home#some_slow_element')
+  expect { @test_site.home.some_slow_element(wait: 1) }
+    .to raise_error(Capybara::ElementNotFound)
```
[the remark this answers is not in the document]
luke-hill: Not sure here, as the behaviour we are mimicking is 100% Wrapped Behaviour, there is no siteprism logic in this cuke. So I don't think we should be wrapping their errors.

I am aware we might do this in other places in our suite, but only when we are performing physical behaviour. The previous test here used our meta-programmed wait method which we are now removing.

lib/site_prism/element_container.rb
tgaff: We could do a final v2.999 build as an upgrade step with deprecations.

tgaff:
Is master now 3.x?   It seems like we're still going to want one more 2.xx build, right?

luke-hill:
Will make a 2.17 cut with deprecations as a siloed branch.

PR #316, opened 2018-08-28
v2.17 branch

luke-hill:
I'll leave this active as long as is required by others / we feel is relevant. (Probably until a stable 3.0 is cut).

I've deprecated some items, and for simplicities sake removed rubocop sign-off. This is cut from the stable 2.16 HEAD revision (So is 12 repo commits behind master as of writing)

Each commit is an individual change / tweak to release as and when we want.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,16 +1,14 @@
 <!-- #291 - Jul 20th - Last update to this document -->
-## [Unreleased]
+## [2.17] - Unreleased
 ### Removed
-- N/A
-
-### Added
-- N/A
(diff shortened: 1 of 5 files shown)


luke-hill:
@tgaff @ineverov Can I get a review on this Standalone branch (It will form the basis of a small bump and a 2.17 release)

Once this is released we'll just use this as a basis if further tweaks are needed. But the plan is still 3.0 development. Reckon I'll get another couple of weeks of dev work in and then think about cutting an alpha version.

luke-hill:
Closing PR as 2.17 has been cut, made into an RC and published to rubygems. Now onwards with 3.0 alpha!

Issue #315, opened 2018-08-23
Refactor / Code-Tidy: Naming of DSL items in pages/section sample files


luke-hill:
The final of a few small issues I'm tagging as "Good for newcomers"

Issue: In general the names of things in the cucumber suite isn't good. We have things that are called `removing_links` e.t.c. and on the whole they should have more semantic names

Pitfalls to watch out for
- Once renaming these, you'll need to check you've hit every occurrence. So long as you have a good IDE you'll cover all of these
- For bonus points see if you can also fix up the file-names for Classes, a filename for a single class should be identical (in snakecase), to the class name.

Feel free to ask if there are any issues.

PR #323, opened 2018-09-09
Initial bout of name fixes

luke-hill:
Initial first attempt pass at fixing up https://github.com/natritmeyer/site_prism/issues/315
Still a bit of work to do.


test_site/pages/home.rb
```diff
@@ -8,59 +8,44 @@ class Home < SitePrism::Page
...
   element :welcome_message, 'body > span'
   element :go_button, '[value="Go!"]'
-  element :link_to_search_page, :xpath, '//p[2]/a'
-  # Below element takes just over 2 seconds to appear
-  element :some_slow_element, :xpath, '//a[@class="slow"]'
-  element :invisible_element, 'input.invisible'
-  element :shy_element, 'input#will_become_visible'
-  element :retiring_element, 'input#will_become_invisible'
+  element :search_page_link, :xpath, '//p[2]/a'
+  element :slow_element, :xpath, '//a[@class="slow"]'
```
tgaff: The comment might actually be useful to keep.

test_site/pages/home.rb
```diff
@@ -8,59 +8,44 @@ class Home < SitePrism::Page
...
-  # elements groups
-  elements :removing_links, '#link_container_will_become_nonexistent > a'
-  elements :nonexistent_elements, 'input#nonexistent'
-  elements :welcome_headers, :xpath, '//h3'
-  elements :welcome_messages, :xpath, '//span'
+  # element groups
+  elements :removing_links, '#removing_link_div > a'
+  elements :nonexistents, 'input#nonexistent'
   elements :rows, 'td'
   # note - actually only one of the below
```
tgaff: this seems like maybe a bug.... should there be more than one?

test_site/pages/home.rb
luke-hill: This comment was incorrect, so removed.

luke-hill:
@tgaff @ineverov My html / js isn't up to scratch, so I googled some tips.

I "think" I've written the js functions in a slightly more efficient way, added a 2nd link (Which we never had enumerable slow links before), and amended some timings to make the suite run even quicker.

Let me know if you think this is ok. I've cut 3.0.beta this morning so this will just hit master and the full 3.0 release.

PR #326, opened 2018-09-25
Error standardisation and Initial Logger

luke-hill:
There's a couple of items left to tweak. But on the whole this is a good first stab at the logger introduction and error tidy.

Note this implementation only makes use of the Logger @ Ruby2.2 (I decided not to add backwards compatibility, as soon we will be Ruby 2.3+ for the suite, once Capybara 3 is the default)

Ping @tgaff @mnohai-mdsol


spec/logger_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        SitePrism.logger.warn('WARN')
+      end
+
+      expect(log_messages.lines).to eq(3)
+    end
+  end
+
+  def capture_stdout
+    original_stdout = $stdout
+    $stdout = fake = StringIO.new
```
mnohai-mdsol: rubocop didn't yell at you for this?

luke-hill: Not gonna lie, I copied this from SO.

I "think" that parallel assignment (Which this is), is ok when it is a unitary assignment, i.e. this creates an object duplicate.

I think rubocop complains when you do `a, b = 1, 2`

No idea if this code method block is optimal, but it seems to work? Is the method ok?

luke-hill:
Re-ping @tgaff / @ineverov if you have time to review this. Fixes up some more stuff for the v3 launch. Nearly there now!

PR #328, opened 2018-10-01
Simplecov config

luke-hill:
Use centralised location for all config options (Not many at the moment)

Enforce a minimum pass-rate of 92% and no decrease in coverage (This will be upped in time)


luke-hill:
ping @tgaff

PR #330, opened 2018-10-02
Bumped gemspec

luke-hill:
- Allow Capybara in any v3 version
- Enforce addressable 2.5+ because ... that's 2 years old, and I'm not sure on the code being consumed in SP for it.
- Bump minimum rubocop (Now it's fixed)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,9 +7,6 @@ Documentation:
 AccessorMethodName:
   Enabled: false
 
-Style/SafeNavigation:
-  Enabled: false
-
 Style/RegexpLiteral:
   Enabled: false
 
--- a/low_spec.gemfile
+++ b/low_spec.gemfile
@@ -2,7 +2,6 @@
 
 source 'https://rubygems.org'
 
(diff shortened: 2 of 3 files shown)


luke-hill:
Just dotting the t's @ineverov

matthewrudy:
@luke-hill interested to see the capybara constraint loosened

but as a head-up,
it's a good idea to set a max for rubocop,
as they always add in new default cops,

eg. these tests are failing because they added a new cop `Layout/EmptyLineAfterGuardClause`

Currently in my main project I specify `~> 0.59.0` then there won't be any surprise failures when I `bundle update`

luke-hill:
Yeh cheers for that. I think I'm going to revert it back to what it was. Because I've raised an issue on their tracker regarding the _memoized method.

For now having it at 0.58 is still reasonably good.

matthewrudy:
Yeah, I saw that one and thought it was funny.

```
Memoized variable @_load_validations does not match method name _load_validations. Use @_load_validations instead.
```

Thanks.

PR #331, opened 2018-10-03
Initial sample htm refactors

luke-hill:
Slow / Vanishing Pages added.

Next step is to integrate these in and put the home / section_experiments page on a massive diet!

Then from there hopefully we're pretty much sorted.


luke-hill:
ping @mnohai-mdsol / @tgaff

mnohai-mdsol:
LGTM, nice job

PR #333, opened 2018-10-05
Minor grammar tweaks

luke-hill:


lib/site_prism.rb
```diff
@@ -13,7 +13,7 @@ module SitePrism
   autoload :Waiter, 'site_prism/waiter'
 
   class << self
-    attr_accessor :enable_logging
+    attr_writer :enable_logging
```
mnohai-mdsol: You can actually make this a ```attr_reader```

.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.2
\ No newline at end of file
+ruby-2.2
```
mnohai-mdsol: Support for Ruby 2.2 has ended, maybe set the minimum to 2.3?

https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

lib/site_prism.rb
luke-hill: This has a separate reader below and needs to be a writer

PR #346, opened 2018-12-12
Eliminate cause of warnings

### LINKED ISSUE #345: Bundle install requires ruby 2.3
# Issue to raise for SitePrism framework
I am working on a PR for Site Prism and encountered this error:

    [nate@Nates-MBP(eliminate_cause_of_warnings) ~/projects/ruby/site_prism]$ bundle install
Fetching gem metadata from https://rubygems.org/........
xpath-3.2.0 requires ruby version >= 2.3, which is incompatible with the current version, ruby 2.2.10p489

## Environment
- This is on osx, but is presumably universal?

Mainly I want to check if there are objections to bumping the .ruby-version to 2.3, given that latest stable is 2.5, this seems like it shouldn't be an issue, but I'm also not an open source project maintainer...

The alternative is to dig through the dependencies and see what version of what gem needs xpath, and at what point it could be restricted in version to not need ruby 2.3.

menge101:
I have a project that makes extensive use of site_prism.  It is currently running on ruby-2.5.5.

I receive tons of warnings of this form:

/Users/nate/.rvm/gems/ruby-2.5.1/gems/site_prism-3.0/lib/site_prism/page.rb:34: warning: instance variable @page not initialized

And for reference $-W is at 1.  This is not with verbose logging on, this is with default logging level set.

I did update to ruby 2.3.0, and ran all tests successfully.  The update to 2.3 was only due to it being the easiest path to bundle installing.

closes #345

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -21,7 +21,7 @@ module SitePrism
       end
 
       def url_matcher
-        @url_matcher || url
+        @url_matcher ||= url
       end
     end
 
@@ -31,7 +31,7 @@ module SitePrism
     # in order, with the default "displayed?" validation being ran _FIRST_
 
     def page
-      @page || Capybara.current_session
+      @page ||= Capybara.current_session
     end
 
     # Loads the page.


.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.2
+ruby-2.3.0
```
luke-hill: Couple of quick things here. Firstly the way rvm and rbenv parse this file means you can be more lax with it (so 2.3 would be preferable)

Also this should mirror the minimum ruby version of the gem (And what the codebase is developed against)

menge101: Certainly, with you guidance on my problem bundling installing, i've cleared that out and am now running on ruby-2.2.

I have already pushed that up.

menge101:
@luke-hill requested changes made and CI is passing.

luke-hill:
Thanks for the contribution

PR #347, opened 2018-12-14
Updated readme to remove wait_for

TheSpartan1980:
Using `wait_for` now generates undefined method `wait_for_<element>' for #<HomePage:0x05bae758>` as the method was removed in 3.0.beta changes made by @luke-hill 👍


luke-hill:
Good spot. What a Warrior!

PR #348, opened 2018-12-20
Item Mapping Refactor

luke-hill:
Begin to leverage the vast quantity of info we know during the DSL build phase when a Thread is started.

This gives us the ability to add more properties if we see fit, but also begin to leverage the ability to start working on the feature requests to extend `#all_there?`

As it's all private stuff, no extra tests added.

cc/ @ineverov / @tgaff to give it a once over if you've got time.


luke-hill:
Re-ping @tgaff / @ineverov - Fixed up some complaints. Will leave this another week then land it.

PR #361, opened 2019-01-28
Fix page inadvertently caching

### LINKED ISSUE #360: Can't reuse SitePrism::Page objects across multiple Capybara::Sessions
## Environment
- Operating System: Ubuntu Trusty (14.04) 
- Gem versions: site_prism 3.0.1, capybara 3.12.0
- Test framework: RSpec
- Driver info: selenium-webdriver gem 3.11.0, docker-selenium [hub@latest](https://hub.docker.com/r/selenium/hub/tags), docker-selenium [node-chrome@3.13](https://hub.docker.com/r/selenium/node-chrome/tags) 

## Expected Behavior
Reusing a single `SitePrism::Page` across multiple `Capybara::Session`s performs actions in the correct browser instance

## Actual Behavior
SitePrism::Page's `@page` is bound to the first Capybara::Session where the page object was used.

## Steps to reproduce
This script assumes that you have chromedriver installed and on your $PATH.
```ruby
require 'capybara/dsl'
require 'selenium-webdriver'
require 'site_prism'

Capybara.current_driver = :selenium_chrome

class Google < SitePrism::Page
  set_url 'https://www.google.com'
  element :privacy, '.Fx4vi', text: 'Privacy'
end

google = Google.new
google.load

Capybara.using_session('session2') do
  google.load
  google.privacy.click
end

raise "I'm on the privacy page, which should be in the other browser" unless google.displayed?
```

luke-hill:
Should Close #360

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -25,13 +25,12 @@ module SitePrism
       end
     end
 
-    # When instantiating the page. A default validation will be added to all
-    # validations you define and add to the Page Class.
-    # When calling #load, all of the validations that are set will be ran
-    # in order, with the default "displayed?" validation being ran _FIRST_
-
     def page
-      @page ||= Capybara.current_session
+      if defined?(@page)
+        @page
+      else
+        Capybara.current_session
+      end
     end
 
     # Loads the page.
@@ -41,6 +40,8 @@ module SitePrism
     #
     # Executes the block, if given.
     # Runs load validations on the page, unless input is a string
+    #
+    # When calling #load, all the validations that are set will be ran in order
     def load(expansion_or_html = {}, &block)
       self.loaded = false
 


luke-hill:
Can you both try this branch on a small side project and check it works as you would intend.
@mresnickmdlive - This shouldn't now leak across sessions
@menge101 - This should still not throw warnings (Note I think your warnings were being triggered because you were launching the ruby interpreter in verbose mode - There are ways it can launch in verbose even if you don't specify it to.)

mresnickmdlive:
@luke-hill Thanks for the quick response and the fix, it's working just fine for me on this branch! I was going to submit a similar PR but I wasn't sure if there were other Capybara objects that SitePrism was storing at the section or element level. I can try adding a Cucumber feature for this if you'd like.

PR #363, opened 2019-02-08
Loadable Checks - Fixes and Tweaks

luke-hill:
Optimise Load Validations to run slightly faster
Ensure checks that were documented take place
Add additional unit / feature tests to cover previously untested code and new code
Fix up bugs (`#loaded` not updating, Validations not being called when calls made without blocks)

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -3,6 +3,7 @@
 require 'site_prism/loadable'
 
 module SitePrism
+  # rubocop:disable Metrics/ClassLength
   class Page
     include Capybara::DSL
     include ElementChecker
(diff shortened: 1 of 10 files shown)


luke-hill:
@ineverov @mnohai-mdsol if you've got time do you want to take a look at this collection of bugfixes / optimisations

lib/site_prism/page.rb
```diff
@@ -140,5 +146,19 @@ def url_matches_by_template?(expected_mappings)
     def matcher_template
       @matcher_template ||= AddressableUrlMatcher.new(url_matcher)
     end
+
+    def load_html_string(string)
```
mnohai-mdsol: these should be private methods I think

luke-hill: They are..?

mnohai-mdsol: ahh, I missed it, lol

mnohai-mdsol:
LGTM

Issue #181, opened 2017-04-13
create_visibility_waiter and similar methods do not combine find_args and runtime_args correctly


koendehondt:
Context in which I found the issue: site_prism 2.9, capybara 2.10.2.

I like to use this:
```
element  :page_title, ".page_title", text: "My page title"
```
and use the element in a visibility waiter:
```
wait_until_page_title_visible
```

That does not work correctly. Capybara shows a warning:
```
gems/capybara-2.10.2/lib/capybara/queries/selector_query.rb:26:in `initialize' :
Unused parameters passed to Capybara::Queries::SelectorQuery : [{:text=>"My page title"}]
```
site_prism's `create_visibility_waiter` does not combine the `find_args`, the `runtime_args`, and `visible: true` correctly. Here is the method:
```
def create_visibility_waiter(element_name, *find_args)
  method_name = "wait_until_#{element_name}_visible"
  create_helper_method method_name, *find_args do
    define_method method_name do |timeout = Waiter.default_wait_time, *runtime_args|
      Timeout.timeout timeout, SitePrism::TimeOutWaitingForElementVisibility do
        Capybara.using_wait_time 0 do
          sleep 0.05 until element_exists?(*find_args, *runtime_args, visible: true)
        end
      end
    end
  end
end
```
In the given example, `create_visibility_waiter` is called with these `find_args`:
```
[".page_title", {:text=>"My page title"}]
```
When `wait_until_page_title_visible` is called, the `runtime_args` are:
```
[]
```
`wait_until_page_title_visible` calls `element_exists?` with these arguments:
```
[".page_title", {:text=>"My page title"}, {:visible=>true}]
```
`element_exists?` sends the arguments to Capybara's `has_selector?`, but the argument list is not supported.

The `find_args`, the `runtime_args`, and `visible: true` should be combined differently, so that the correct argument list is passed to `element_exists?`. Something like this:
```
def create_visibility_waiter(element_name, *find_args)
  method_name = "wait_until_#{element_name}_visible"
  create_helper_method method_name, *find_args do
    define_method method_name do |timeout = Waiter.default_wait_time, *runtime_args|
      Timeout.timeout timeout, SitePrism::TimeOutWaitingForElementVisibility do
        Capybara.using_wait_time 0 do
          find_options = find_args.last.is_a?(Hash) ? find_args.pop : {}
          runtime_options = runtime_args.last.is_a?(Hash) ? runtime_args.pop : {}
[...]

PR #367, opened 2019-02-19
add upgrading info for wait_until methods

tgaff:
Hi Luke, I upgraded one repo to v3 today and encountered an unexpected set of warnings in the output like: `Unused parameters passed to Capybara::Queries::SelectorQuery : [2]`.

I initially thought I'd perhaps reproduced #181 but much of that code had already been refactored.  In the end I realized this is actually an API change - looking through 2.x/3.x history it looks like it occurred at the 3.x upgrade point which makes sense.  

It's not listed in the UPGRADING doc so adding it now.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -54,3 +54,24 @@ Previously `site_prism` (As of `2.17.1`), had 3 configuration options. These wer
 These have all been removed with the `3.0` release. Implicit Waiting is
 controlled / modifiable either in-line for each method-call, or you can set the default
 timeout by re-configuring `Capybara.default_max_wait_time`.
+
+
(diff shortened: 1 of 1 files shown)


tgaff:
`invalid argument: can't kill an exited process (Selenium::WebDriver::Error::UnknownError)`

That test failure doesn't make sense...(especially for a doc change).  Maybe you can re-start the tests.

luke-hill:
This is fine to merge as is. Travis has been playing up with 2.6 and head versions for a while (Most people will be using this gem on 2.2 - 2.5 versions anyways)

Good to have you back. Thanks for the contribution!

PR #368, opened 2019-02-20
Update travis to use Xenial, fix some docs

luke-hill:


luke-hill:
@RustyNail I've put a post on the Travis forums. But given you did a lot of our work on this, if you've got any ideas let me know.

PR #375, opened 2019-03-11
iFrame tweaks

luke-hill:
Add a bit of logging, Update some comments, and add some additional helper methods for iFrames that have been missing for years...?

### DIFF
--- a/lib/site_prism/dsl.rb
+++ b/lib/site_prism/dsl.rb
@@ -66,8 +66,7 @@ module SitePrism
       options[:wait] = wait_time unless wait_key_present?(options)
     end
 
-    # True if the +wait+ key is present in the options hash.
-    #
+    # Detect if the +wait+ key is present in the options hash.
     # Note that setting it to to false or 0, still will return true here.
     def wait_key_present?(options)
       options.key?(:wait)
@@ -228,25 +227,29 @@ module SitePrism
       end
 
       def deduce_iframe_scope_find_args(args)
+        warn_on_invalid_selector_input(args)
         case args[0]
-        when Integer
-          [args[0]]
-        when String
-          [:css, args[0]]
-        else
-          args
+        when Integer then [args[0]]
+        when String  then [:css, args[0]]
+        else               args
         end
       end
 
       def deduce_iframe_element_find_args(args)
+        warn_on_invalid_selector_input(args)
         case args[0]
-        when Integer
-          "iframe:nth-of-type(#{args[0] + 1})"
-        when String
-          [:css, args[0]]
-        else
-          args
+        when Integer then "iframe:nth-of-type(#{args[0] + 1})"
+        when String  then [:css, args[0]]
+        else               args
         end
+      end
+
+      def warn_on_invalid_selector_input(args)
+        return unless args[0].is_a?(String) || !args[0].start_with?('/', './')
+
+        msg = 'The arguments passed in look like xpath. Check your locators.'
+        SitePrism.logger.warn(msg)
+        SitePrism.logger.debug("Current default locator: #{Capybara.default_selector}")
       end
 
       def extract_section_options(args, &block)
--- a/lib/site_prism/dsl.rb
+++ b/lib/site_prism/dsl.rb
@@ -123,13 +123,13 @@ module SitePrism
       def iframe(name, klass, *args)
         element_find_args = deduce_iframe_element_find_args(args)
         scope_find_args = deduce_iframe_scope_find_args(args)
-        map_item(:iframe, name)
-        add_iframe_helper_methods(name, *element_find_args)
-        define_method(name) do |&block|
-          raise MissingBlockError unless block
 
-          within_frame(*scope_find_args) do
-            block.call(klass.new)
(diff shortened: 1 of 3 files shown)


luke-hill:
ping @tgaff / @mnohai-mdsol

lib/site_prism/dsl.rb
```diff
@@ -228,27 +222,31 @@ def create_error_method(name)
...
-          [:css, args[0]]
-        else
-          args
+        when Integer then "iframe:nth-of-type(#{args[0] + 1})"
+        when String  then [:css, args[0]]
+        else args
         end
       end
 
+      def warn_on_invalid_selector_input(args)
```
mnohai-mdsol: Actually, why can't we use xpath? As long as you have something like:

```ruby
page.driver.browser.switch_to.frame
```
You should be able to use the selector of choice.

mnohai-mdsol: I see, yes, they should be defining ```:xpath``` as the selector if they choose to need it.

mnohai-mdsol: I would merge it for you, but I don't have access to.

lib/site_prism/dsl.rb
```diff
@@ -145,13 +144,13 @@ def build(type, name, *find_args)
...
         else
-          map_item!(type, name)
+          map_item(type, name)
           yield
         end
         add_helper_methods(name, *find_args)
       end
 
-      def map_item!(type, name)
+      def map_item(type, name)
```
ineverov: I know it's not related to this PR, but why mapped_items is an array of hashes `type -> name`. Any use case?
IMO, it would be better to have mapped_items to be something like, `element: [:name1, :name2, :name3], section: [:section1, section2, section3]`

lib/site_prism/dsl.rb
```diff
@@ -228,27 +222,31 @@ def create_error_method(name)
...
-        else
-          args
+        when Integer then "iframe:nth-of-type(#{args[0] + 1})"
+        when String  then [:css, args[0]]
+        else args
         end
       end
 
+      def warn_on_invalid_selector_input(args)
+        return unless !args[0].is_a?(String) || !args[0].start_with?('/', './')
```
ineverov: It might be re-written like this
`return if args[0].is_a?(String) && args[0].start_with?('/', './')`
Can `args[0]` be `nil` by mistake? We should provide a meaningful message than

And it will warn if `args[0]` is not a string, which is not expected behavior. 

The condition should be something like
`return if !args[0].is_a?(String) || args[0].start_with?('/', './')`

lib/site_prism/dsl.rb
luke-hill: This was something I changed a bit ago, and it's going to help the framework of recursion for all_there.

It could be in the form you've suggested sure. If you wanna raise a PR go for it.

Currently these are only used for `all_there?` and `mapped_items` checks

lib/site_prism/dsl.rb
luke-hill: If it's `nil` it gets caught in the `build` phase and will create error methods for everything and crazy lazily. So we're fine with that 2nd condition (I've fixed up the first bit as per DeMorgans Law)

luke-hill:
Merging in. Issue with Pulling 2.6 images (Master build should fix this)

lib/site_prism/dsl.rb
```diff
@@ -228,27 +222,35 @@ def create_error_method(name)
...
+      def warn_on_invalid_selector_input(args)
+        return if looks_like_xpath?(args[0])
+
+        msg = 'The arguments passed in look like xpath. Check your locators.'
+        SitePrism.logger.warn(msg)
+        SitePrism.logger.debug("Default locator: #{Capybara.default_selector}")
+      end
+
+      def looks_like_xpath?(arg)
+        arg.is_a?(String) && arg.start_with?('/', './')
```
ineverov: I have a feeling there is still an error here.
condition above should return if it is not-xpath-like string or another type (an integer for example)
so we assert it is a string with `arg.is_a?(String)` if it is false we should exit so let's write it down
`return if !arg.is_a?(String)`, next, if it is a string we check if it doesn't start with a slash `arg.start_with?('/', './')`. so the final condition is `return if !arg.is_a?(String) || arg.is_a?(String) && arg.start_with?('/', './')` or after simplifications `!A || A && B` => `!(A && (!A || !B))` => `!( A && !A || A && !B)` => `!(A && !B)` => `!A || B` => `return if !arg.is_a?(String) || arg.start_with?('/', './')`

lib/site_prism/dsl.rb
```diff
@@ -228,27 +222,35 @@ def create_error_method(name)
...
-        else
-          args
+        when Integer then "iframe:nth-of-type(#{args[0] + 1})"
+        when String  then [:css, args[0]]
+        else args
         end
       end
 
+      def warn_on_invalid_selector_input(args)
+        return if looks_like_xpath?(args[0])
```
ineverov: I have a feeling there is still an error here. There should be `unless` here

PR #381, opened 2019-04-03
Refactor: Create #wait_until_displayed? - standardise #displayed? method

TheSpartan1980:
@luke-hill
**First pass at moving the code to the Waiter**
Moved logic into the Waiter
Added url_matcher as a parameter to the wait_until_displayed? method
Pass url_matches? as a block however found that it always returns true and breaks two unit tests

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -65,7 +65,7 @@ module SitePrism
 
     def displayed?(*args)
       begin
-        Waiter.wait_until_displayed?(*args, url_matcher) { url_matches? } 
+        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(*args) } 
       rescue SitePrism::TimeoutError
         false
       end
@@ -121,7 +121,8 @@ module SitePrism
       matcher_template.mappings(page.current_url)
     end
 
-    def url_matches?(expected_mappings = {})
+    def url_matches?(*args)
+      expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
       if url_matcher.is_a?(Regexp)
         url_matches_by_regexp?
       elsif url_matcher.respond_to?(:to_str)
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -64,8 +64,9 @@ module SitePrism
     end
 
     def displayed?(*args)
+      expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
       begin
-        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(*args) } 
+        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(expected_mappings) } 
       rescue SitePrism::TimeoutError
         false
       end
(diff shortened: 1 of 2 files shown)


lib/site_prism/waiter.rb
```diff
@@ -14,5 +14,12 @@ def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
 
       raise SitePrism::TimeoutError, "Timed out after #{wait_time}s."
     end
+
+    def self.wait_until_displayed?(*args, url, &url_matches)
```
luke-hill: This won't be a predicate, so lets drop the `?`

lib/site_prism/page.rb
```diff
@@ -64,12 +64,8 @@ def load(expansion_or_html = {}, &block)
     end
 
     def displayed?(*args)
-      expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
-      seconds = !args.empty? ? args.first : Capybara.default_max_wait_time
-      raise SitePrism::NoUrlMatcherForPageError unless url_matcher
-
       begin
-        Waiter.wait_until_true(seconds) { url_matches?(expected_mappings) }
+        Waiter.wait_until_displayed?(*args, url_matcher) { url_matches? }
```
luke-hill: lets make this method operate like a predicate, drop all waiting logic

lib/site_prism/waiter.rb
TheSpartan1980: Removed '?' from the method

lib/site_prism/page.rb
TheSpartan1980: @luke-hill I've pushed a new commit, I think it's a bit ugly but all the current tests are now passing which is a small comfort

lib/site_prism/page.rb
```diff
@@ -64,12 +64,8 @@ def load(expansion_or_html = {}, &block)
     end
 
     def displayed?(*args)
-      expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
-      seconds = !args.empty? ? args.first : Capybara.default_max_wait_time
-      raise SitePrism::NoUrlMatcherForPageError unless url_matcher
-
       begin
-        Waiter.wait_until_true(seconds) { url_matches?(expected_mappings) }
+        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(*args) }
```
luke-hill: passing the extra var through is un-needed, everything is auto-loaded from the raw DSL.

If you find this causes issues, require the requisite file in the Waiter class

lib/site_prism/page.rb
```diff
@@ -125,7 +121,8 @@ def template_backed_matches
       matcher_template.mappings(page.current_url)
     end
 
-    def url_matches?(expected_mappings = {})
+    def url_matches?(*args)
+      expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
```
luke-hill: This feels like a duplication.

If this is the only usage, then lets dry up the other call, otherwise, if you're not sure, remove this and convert the PR into reviewable.

lib/site_prism/page.rb
TheSpartan1980: @luke-hill `url_matches?` is `private` within the page class so it causes an issue in that respect

luke-hill: But it will always be calling it from within another method, so it shouldn't cause any issues. If loading is a problem, require the file.

TheSpartan1980:
@luke-hill I seem to be chasing my tail with this one so I returned some of the logic back to displayed? additionally when `url_matcher` is call within page.rb for `spec/site_prism/page.rb:308` it returns a value. However when it's call from within `waiter.rb` nil is returned. Additionally I couldn't find a nice way to abstract out `url_matches?`

PR #383, opened 2019-04-16
Allow to pass with_validations: false to the load command

### LINKED ISSUE #382: Ability to skip load validation
Hello,

I would like to optionally skip load validations. For example I'm testing a redirect flow where in some cases the page I try to visit redirects to another page first because the user needs to perform an action.

1. go to `/app`
2. redirect to `/app/accept_gdpr`
3. user accepts and we are back on `/app`

When step 2 is already done we instantly go to `/app`, with site prism 3.1 I get the following error on step 1: `Expected http://127.0.0.1/app/accept_gdpr to match /app but it did not.`

Which is obvious since we navigate to `/app` but get redirected to `/app/accept_gdpr`. So my question is how can I temporarily disable the load validation?

JanStevens:
Fixes #382 

The load method signature did change but I think thats fine, I updated the page_spec bug not sure if there needs to be any other spec updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -1345,6 +1345,20 @@ Load validations may be defined on `SitePrism::Page` and `SitePrism::Section`
 classes (herein referred to as `Loadables`) and are evaluated against an
 instance of the class when created.
 
+### Skipping load Validations
+
+Defined load validations can be skipped for one `load` call by 
+passing in `with_validations: false`.
+
+```ruby
+it 'loads the page without validations' do
+  some_page.load(with_validations: false)
+  some_page.loaded?    #=> true if/when all load validations pass
+  another_page.loaded? #=> false if any load validations fail
+  another_page.load_error #=> A string error message if one was supplied by the failing load validation, or nil
+end
+```
+
 ### Load Validation Inheritance and Execution Order
 
 Any number of load validations may be defined on a Loadable and they are
--- a/README.md
+++ b/README.md
@@ -1353,9 +1353,7 @@ passing in `with_validations: false`.
 ```ruby
 it 'loads the page without validations' do
   some_page.load(with_validations: false)
-  some_page.loaded?    #=> true if/when all load validations pass
-  another_page.loaded? #=> false if any load validations fail
-  another_page.load_error #=> A string error message if one was supplied by the failing load validation, or nil
+  some_page.loaded?    #=> true unless something has gone wrong
 end
 ```
 
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -35,6 +35,8 @@ module SitePrism
     end
 
     # Loads the page.
+    # @param html HTML string to load
+    # @param with_validations
     # @param expansion_or_html
     # @param block [&block] An optional block to run once the page is loaded.
     # The page will yield the block if defined.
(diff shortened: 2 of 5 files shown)


lib/site_prism/page.rb
```diff
@@ -43,14 +45,14 @@ def page
     # Runs load validations on the page, unless input is a string
     #
     # When calling #load, all the validations that are set will be ran in order
-    def load(expansion_or_html = {}, &block)
+    def load(html = nil, with_validations: true, **expansion, &block)
```
luke-hill: can we just add the `with_validations` key here so that it's not backwards incompatible please. As this is quite a widely used method.

lib/site_prism/page.rb
```diff
@@ -152,11 +154,13 @@ def load_html_string(string)
       yield self if block_given?
     end
 
-    def load_html_website(html, &block)
+    def load_html_website(html, with_validations, &block)
       expanded_url = url(html)
       raise SitePrism::NoUrlForPageError unless expanded_url
 
       visit expanded_url
+      return yield self unless with_validations
```
luke-hill: This will cause issues as a different yield will be returned dependent on situation. The return value of this method will be yielded later on in the method invocation chain.

spec/site_prism/page_spec.rb
```diff
@@ -168,6 +168,16 @@ def foo?
             .to raise_error(SitePrism::FailedLoadValidationError)
             .with_message('It is not true!')
         end
+
+        context 'when validations are disabled' do
```
luke-hill: can we add this to the positive use case as well

lib/site_prism/page.rb
JanStevens: `def load(expansion_or_html = {}, with_validations: true, &block)` doesn't work since when you pass in `username: 'foobar'` it blows up with `#<ArgumentError: unknown keyword: username>`.

Another way would be to do a `expansion_or_html.fetch(:with_validations, false)` (before `load_html_website`)

The current change is 100% backwards compatible (assuming specs test every possible use case)

JanStevens:
I've added more specs and we also now test the case where you have load validations and there is no block passed (this would indeed not work as you pointed out but no spec failed for it)

lib/site_prism/page.rb
luke-hill: The 2nd option you mentioned, that just assumes that no-one nestles that option inside the hash. I think this is a valid assumption (If you don't want to place it as a config option). Because that expansion_or_html hash is only ever realistically going to have a couple of options max in it.

JanStevens:
I've changed it so we don't change the method signature

luke-hill:
ping @ineverov anything you can think of for this?

luke-hill:
Code-wise this is fine, but can you do the following.
- Add a feature test, you shouldn't need to write any new fixtures. Just an additional scenario. It should live in navigation, and probably should be similar the other load validation tests
- Add a piece of documentation about how this works

README.md
```diff
@@ -1345,6 +1345,20 @@ Load validations may be defined on `SitePrism::Page` and `SitePrism::Section`
...
 
+### Skipping load Validations
+
+Defined load validations can be skipped for one `load` call by 
+passing in `with_validations: false`.
+
+```ruby
+it 'loads the page without validations' do
+  some_page.load(with_validations: false)
+  some_page.loaded?    #=> true if/when all load validations pass
```
luke-hill: Can you alter the comment here, this is actually invalid as it won't run load validations. Just something like. `true` unless something has gone wrong (Keep it vague is fine)

Can we also delete the two lines below.

luke-hill:
Thanks a lot for your contribution :)

PR #384, opened 2019-04-17
Refactor: Create #wait_until_displayed - standardise #displayed? method

TheSpartan1980:
Moved the majority of the logic to the `waiter`
Ensured all specs are still passing
Looked for the opportunity to write additional specs, however it seems the currents specs more than cater for the code.

### DIFF
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -66,7 +66,7 @@ module SitePrism
     def displayed?(*args)
       expected_mappings = args.last.is_a?(::Hash) ? args.pop : {}
       begin
-        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(expected_mappings) } 
+        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(expected_mappings) }
       rescue SitePrism::TimeoutError
         false
       end
--- a/lib/site_prism/page.rb
+++ b/lib/site_prism/page.rb
@@ -65,7 +65,7 @@ module SitePrism
 
     def displayed?(*args)
       begin
-        Waiter.wait_until_displayed?(*args, url_matcher) { url_matches? } 
+        Waiter.wait_until_displayed(*args, url_matcher) { url_matches?(*args) } 
       rescue SitePrism::TimeoutError
         false
       end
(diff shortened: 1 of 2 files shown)


TheSpartan1980:
A BIG thank you! to @luke-hill for his patience and guidance

lib/site_prism/waiter.rb
```diff
@@ -14,5 +14,13 @@ def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
 
       raise SitePrism::TimeoutError, "Timed out after #{wait_time}s."
     end
+
+    def self.wait_until_displayed(*args, url)
```
luke-hill: This method should essentially house 95% of the logic. And doesn't need to leak state from the original method call. Remember both of these are essentially auto-loaded in

lib/site_prism/waiter.rb
```diff
@@ -14,5 +14,13 @@ def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
...
+      args.last.is_a?(::Hash) ? args.pop : {}
+      seconds = !args.empty? ? args.first : Capybara.default_max_wait_time
+      raise SitePrism::NoUrlMatcherForPageError unless url
```
luke-hill: Lets optimise this method call by making this a proper guard clause on line1 of the call

lib/site_prism/waiter.rb
```diff
@@ -14,5 +14,13 @@ def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
 
       raise SitePrism::TimeoutError, "Timed out after #{wait_time}s."
     end
+
+    def self.wait_until_displayed(*args, url)
+      args.last.is_a?(::Hash) ? args.pop : {}
```
luke-hill: lets copy this from the original method

lib/site_prism/page.rb
```diff
@@ -65,11 +65,8 @@ def load(expansion_or_html = {}, &block)
 
     def displayed?(*args)
```
luke-hill: lets rewrite this method as the following.

```
def displayed?(*args)
  Waiter.wait_until_displayed(*args) { url_matches?(expected_mappings) }
rescue SitePrism::TimeoutError
  false
end
```

luke-hill:
@TheSpartan1980 are you able to still tackle this now? Want to maybe get this merged in and cut a v3.2 release as a few people have merged in some new features.

TheSpartan1980:
Hi @luke-hill I think I'm going to park this for now I just can't seem to get my head around it. It seems whenever I make changes pre-existing tests break and then when I refactor it to make the test pass the code is less ideal.

luke-hill:
I'll Cherry-pick your commits and pick it up. As I've got nothing I'm working on currently.

PR #385, opened 2019-04-18
#all_there? - Initial Recursion Logic

luke-hill:
Fully TDD'd, with semi-functional test coverage and logic which partially works.

ping @tgaff for review/help/usage pointers (I know you asked).

NB: This currently is not "fully" compatible with `.expected_items`.


luke-hill:
This now in theory "begins" to use `.expected_items` - But I've not given it anywhere near enough mileage to test this. @tgaff if you're up for giving this a bit of a runabout let me know if it works as you'd expect.

luke-hill:
@faceless7171 @ineverov could you both take a look at this. See if it works / you can spot any issues.

luke-hill:
re-ping @tgaff / @faceless7171 - Could you give me some feedback on this, otherwise I'll probably just leave it be.

faceless7171:
Yes, sure, I will take a look!

luke-hill:
This is now fully up to date with v3.2 of the gem, and has been tidied up and slightly optimised. There's still a chunk more stuff to do. But I think I'm going to get this released sooner rather than later and then work on splitting it out into its own gem, as there's a whole heap of stuff that's not possible with the recursion strategies we've uncovered.

faceless7171:
@luke-hill Sorry was a little bit busy. For the things I can see this is looks good, great job!

Issue #388, opened 2019-04-25
Siteprism hangs when used with timecop


dkniffin:
I've just discovered an issue with interactions between site_prism and timecop. It looks like this is a known issue (#88), which was closed (completely reasonably).

To describe the issue again here: If you are using [timecop](https://github.com/travisjeffery/timecop) with `Timecop.freeze` somewhere in your code (either in the feature/system test, or in another test with leakage), `Waiter.wait_until_true` will hang because `Time.now` does not change.

This actually isn't a bug with Site Prism, but I just spent a few hours tracking this down, so I wanted to create an issue, to hopefully avoid this issue for others in the future. At the very least, I'm hoping this issue will come up in a Google search, but I'm also wondering if there's some documentation or messaging we could add that might help as well.

My two thoughts:
1. We could put a bit of code in somewhere that checks for the presence of `Timecop` and if it's there, just output some logging that says something like "fyi, you're using timecop"
2. We could add a "known issues" section to the readme listing this, so people can find it.

luke-hill:
I know previously we've had some `Timeout.timeout` calls we fixed for this, and I know Capybara have got an experimental PR (Thomas has had this for a while), to crash when people call `Timecop.freeze`

As mentioned in the original Issue request, browser integration tests should never have frozen time. Whereas maybe stubbing it in specs is fine (However we do have some time management unit tests also which may have issues with this)

Both of your points are completely valid, and I'm happy to listen to PR's for them. At the moment I'm deep inside a couple of gritty features (Which are on the repo and have been for a while), so I'd prefer to keep my focus on tackling these.

https://github.com/teamcapybara/capybara/pull/2032/files - This is the PR which is being experimented with to prevent time freezes.

Caveats to my acceptance of your above ideas are...
1) What should we do if we encounter frozen time..
  - Crash
  - Logger messages (site_prism has it's own logging API)
2) Should we add timecop as a dependency and monitor it, or perhaps (Not sure if this is possible), ban the usage of timecop with site_prism

I'll also happily listen to a listener-poll inside the `wait_until_true` method that after each tick checks whether the elapsed time was >0 (And if it wasn't throw an error and exit with a `FrozenInTimeError`

Hopefully that's given you a few nuggets of wisdom to debate. But at the moment I wouldn't be able to commit to coding it. If you or others watching are then I'm all for extra functionality / guards.

dkniffin:
[...]

PR #389, opened 2019-04-25
Add SitePrism::FrozenInTimeError to catch frozen time issues

dkniffin:
#388

### DIFF
--- a/lib/site_prism/waiter.rb
+++ b/lib/site_prism/waiter.rb
@@ -2,17 +2,33 @@
 
 module SitePrism
   class Waiter
-    def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
-      start_time = Time.now
+    class << self
+      def wait_until_true(wait_time = Capybara.default_max_wait_time)
+        start_time = Time.now
 
-      loop do
-        return true if yield
-        break if Time.now - start_time > wait_time
+        loop do
+          return true if yield
+          break if Time.now - start_time > wait_time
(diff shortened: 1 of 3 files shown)


lib/site_prism/waiter.rb
```diff
@@ -10,9 +10,23 @@ def self.wait_until_true(wait_time = Capybara.default_max_wait_time)
...
+
+        check_for_time_stopped!
       end
 
       raise SitePrism::TimeoutError, "Timed out after #{wait_time}s."
     end
+
+    private
+
+    def check_for_time_stopped!
```
luke-hill: this needs to take the start_time as an argument I think.

spec/site_prism/waiter_spec.rb
```diff
@@ -26,5 +26,22 @@
 
       expect(duration).to be_within(0.1).of(timeout)
     end
+
+    context 'when time is frozen' do
+      before do
```
luke-hill: this could probably be simplified to `allow(Time).to receive(:now).and_return { Time.new(2019,4,25) }` to save you having to open the class up - Might be a bit shorter as well

dkniffin:
@luke-hill Alright, fixed all that up, and tests are passing. Let me know if there's anything else needed here.

luke-hill:
Top stuff, well done on the quick turnaround too. Thanks for the contribution, it's greatly appreciated.

PR #390, opened 2019-04-30
Added support for negated existence matchers in RSpec

hoffi:
This adds support for negated existence matchers in RSpec. E.g.:
```ruby
  element :heading, 'h1'
end
expect(page).not_to have_heading # => Calls now #has_no_heading?
```

This also worked before however it used the `#has_<element>?` methods for that case. This caused capybara to wait until the element **is** visible. But as in the example above i expect the element to be **not** visible this always waits until the max wait timeout is reached.

This change automatically calls the `#has_no_<element>?` method when the RSpec expectation is negated.

### DIFF
--- a/spec/site_prism/element_spec.rb
+++ b/spec/site_prism/element_spec.rb
@@ -16,10 +16,13 @@ describe 'Element' do
     it { is_expected.to respond_to(:wait_until_element_one_visible) }
     it { is_expected.to respond_to(:wait_until_element_one_invisible) }
 
+    it 'supports rspec existence matchers' do
+      expect(subject).to have_element_one
+    end
+
     it 'supports negated rspec existence matchers' do
-      expect(subject).not_to receive(:has_element_one?)
-      expect(subject).to receive(:has_no_element_one?).and_return true
-      expect(subject).not_to have_element_one
+      expect(subject).to receive(:has_no_element_two?).once.and_call_original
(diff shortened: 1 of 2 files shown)


spec/site_prism/element_spec.rb
```diff
@@ -16,6 +16,12 @@
     it { is_expected.to respond_to(:wait_until_element_one_visible) }
     it { is_expected.to respond_to(:wait_until_element_one_invisible) }
 
+    it 'supports negated rspec existence matchers' do
```
luke-hill: Can we re-write these using the element stubs and turn them into 1liners, and add both options i.e.
```
it 'supports rspec existence matchers'
  expect(subject).to have_element_two # I think
end
```

hoffi: I would like to keep the one expectation that the negated matcher is called. Would that be ok?

```ruby
it 'supports rspec existence matchers' do                                
  expect(subject).to have_element_one                                    
end                                                                      
                                                                         
it 'supports negated rspec existence matchers' do                        
  expect(subject).to receive(:has_no_element_two?).once.and_call_original
  expect(subject).not_to have_element_two                                
end                                                                      
```

hoffi: Added in a74be32

luke-hill:
Merged. Thankyou for your contribution

PR #391, opened 2019-05-08
Refactor #displayed? into Waiter and regular boolean

luke-hill:
Cherry-picked and derived from https://github.com/natritmeyer/site_prism/pull/384

### DIFF
--- a/.simplecov
+++ b/.simplecov
@@ -6,4 +6,4 @@ SimpleCov.start do
   add_group 'Code', 'lib'
 end
 
-SimpleCov.minimum_coverage 97
+SimpleCov.minimum_coverage 98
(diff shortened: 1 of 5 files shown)


luke-hill:
@TheSpartan1980 this will be released in the next version of site_prism (3.2) - Thanks for your help with the original work!

TheSpartan1980:
@luke-hill great work we're really lucky and grateful to have you as the lead maintainer. No idea why I became so fixated with having the logic in the waiter, next time I'll do better 😊

luke-hill:
Yeh I think the problem is that you're migrating from class methods to instance methods, so the inheritance chain is screwy (include vs extend).

Also this allows people to just use a drop-in-replacement, alongside Load Validations and such

PR #394, opened 2019-05-24
Remove outdated RSpec syntax

andyw8:
The cardinality matchers were removed in RSpec 3.0.0.beta1. They're available using https://github.com/rspec/rspec-collection_matchers, but since that's not available by default, it might cause confusion to mention it here.


luke-hill:
Thanks for your contribution!

